PR #1, opened 2021-11-13
Fix MUX toggling issue 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

Fix the issue that randomly fails MUX state toggling. 

To be more specific, the issue was when trying to toggle MUX ports from standby to active utilizing cli  "config mux mode standby all", some random ports won't be toggled successfully and will remain standby mode. This issue can be easily reproduced when ICMP_responder is off but not when the ports are in healthy state. 

The cause is when ICMP_responder is off and ToRs stay in unhealthy state, linkmgrd will probe mux state every 300ms[^1]  which will make mux state enter "wait". While mux state equals "wait", linkmgrd won't switch mux sate to user's desiring state[^2]. 

A state machine diagram for this particular scenario can be found under the internal sharepoint folder for Gemini design doc.  

[^1]: This interval will be multiplied by a backoff factor if the returned state is unknown. 
[^2]: Instead, it will do a mux state probe. 

Signed-off-by: Jing Zhang <zhangjing@microsoft.com>

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)

### Back port request
- [ ] 201911

We want to make sure users are able to toggle from standby to active even when icmp responder is not working, without manual retries.  

#### How did you do it?

Cache the MUX mode configuration change, execute the change once MUX exits "wait" state. 

Tested it on virtual dual testbeds, unable to reproduce the issue anymore.

### DIFF
--- a/src/link_manager/LinkManagerStateMachine.cpp
+++ b/src/link_manager/LinkManagerStateMachine.cpp
@@ -506,8 +506,13 @@ void LinkManagerStateMachine::handleStateChange(MuxStateEvent &event, mux_state:
         // Verify if state db MUX state matches the driver/current MUX state
         mMuxPortPtr->getMuxState();
 
-        // Handle pening mux mode config change
+        // Handle pending mux mode config change
         if (mPendingMuxModeChange) {
+            MUXLOGINFO(boost::format("%s: Mux state: %s . Execute pending MUX mode config change.") % 
+                mMuxPortConfig.getPortName() %
+                mMuxStateName[ms(mCompositeState)]
+            );
+
             handleMuxConfigNotification(mTargetMuxMode);
             mPendingMuxModeChange = false;
         }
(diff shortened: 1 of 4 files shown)


zjswhhh:
It appears that Tamer set up some pipelines before, I will check

lguohan:
can you add unit test code?

lguohan:
the description is unclear, please clearly describe the problem and how you did to address it.

lguohan (changes requested):
need description and unit test code

yxieca:
@zjswhhh the change looks good. I agree with Guohan, please add unit test.

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -505,6 +505,12 @@ void LinkManagerStateMachine::handleStateChange(MuxStateEvent &event, mux_state:
     if (ms(mCompositeState) != mux_state::MuxState::Wait) {
         // Verify if state db MUX state matches the driver/current MUX state
         mMuxPortPtr->getMuxState();
+
+        // Handle pening mux mode config change
+        if (mPendingMuxModeChange) {
+            handleMuxConfigNotification(mTargetMuxMode);
+            mPendingMuxModeChange = false;
+        }
```
yxieca, line 513: Suggest making above code a helper member function.

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -706,28 +712,41 @@ void LinkManagerStateMachine::handleSwssLinkStateNotification(const link_state::
...
+        ms(mCompositeState) == mux_state::MuxState::Wait) {
+        
+        MUXLOGWARNING(boost::format("%s: Mux mode: %s , mux mode config change is pending. ") % 
+                mMuxPortConfig.getPortName() %
+                mMuxStateName[ms(mCompositeState)]
+            );
+
+        mPendingMuxModeChange = true;
+        mTargetMuxMode = mode;
+    } else {
```
yxieca, line 727: Suggest always record the pending action first. And make the log level notice. (code consistency)

Then call the helper function if the state is right below.

Add a matching notice log in the helper function so that we know when the request was executed.

src/link_manager/LinkManagerStateMachine.cpp
yxieca: I think this check and execution should be placed at the end of the function.

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -505,6 +505,12 @@ void LinkManagerStateMachine::handleStateChange(MuxStateEvent &event, mux_state:
     if (ms(mCompositeState) != mux_state::MuxState::Wait) {
         // Verify if state db MUX state matches the driver/current MUX state
         mMuxPortPtr->getMuxState();
+
+        // Handle pening mux mode config change
```
theasianpianist, line 509: typo

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -706,28 +712,41 @@ void LinkManagerStateMachine::handleSwssLinkStateNotification(const link_state::
...
+        mode != common::MuxPortConfig::Mode::Manual &&
+        ms(mCompositeState) == mux_state::MuxState::Wait) {
+        
+        MUXLOGWARNING(boost::format("%s: Mux mode: %s , mux mode config change is pending. ") % 
+                mMuxPortConfig.getPortName() %
+                mMuxStateName[ms(mCompositeState)]
+            );
+
+        mPendingMuxModeChange = true;
+        mTargetMuxMode = mode;
```
yxieca, line 726: Consider return here. and Indent else block back.

vdahiya12:
@zjswhhh lets also run this change with all DualToR sonic-mgmt tests and see if they pass.

src/link_manager/LinkManagerStateMachine.cpp
zjswhhh: > Consider return here. and Indent else block back.

Updated accordingly.

src/link_manager/LinkManagerStateMachine.cpp
zjswhhh: > typo

Updated accordingly.

zjswhhh:
> need description and unit test code

Added accordingly.

zjswhhh:
> @zjswhhh lets also run this change with all DualToR sonic-mgmt tests and see if they pass.

Will do.

zjswhhh:
> > @zjswhhh lets also run this change with all DualToR sonic-mgmt tests and see if they pass.
> 
> Will do.

@vdahiya12 Confirmed that this change wouldn't break any dualtor test that passed on 20201231 release versions. There are 4 test cases fail consistently even on release versions, we can take a look offline if needed.

vdahiya12:
> > > @zjswhhh lets also run this change with all DualToR sonic-mgmt tests and see if they pass.
> > 
> > 
> > Will do.
> 
> @vdahiya12 Confirmed that this change wouldn't break any dualtor test that passed on 20201231 release versions. There are 4 test cases fail consistently even on release versions, we can take a look offline if needed.
Thanks, did we run both dualtor and dualtor_io ?

lguohan (approved):

lguohan:
when you merge this pr, please use squash merge and also make sure the commit message is properly formatted and have all the information in your pr description.

yxieca (approved):

PR #6, opened 2021-11-23
Force mux state switch to standby if triggered from Cli

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

Fix the timestamp gap between linkmgrd_switch_active_start and  linkmgrd_switch_standby_start events.

When switching from active to standby using ```config mux mode standby``` on the active ToR, the active side will first send a "notification" with COMMAND_SWITCH_ACTIVE to standby ToR, then start to switch itself. The switchover start events are expected to be posted at nearly the same time on two ToRs, but in reality we saw a gap around 100ms between these two events. 

The cause is, when handle mux config notification, mux mode was set to standby from auto. When active ToR completes sending COMMAND_SWITCH_ACTIVE message and starts to switch itself, its mode is not auto anymore, hence [it starts to do mux state probing instead of a force switchover](https://github.com/Azure/sonic-linkmgrd/blob/c11a5763a070a68af048ccd9ca473294f93fb256/src/link_manager/LinkManagerStateMachine.cpp#L346-L366). Hence, cause the delay. 

The fix is a force switch, which makes sense from my point of view as user's desiring state is standby in this scenario. 

Signed-off-by: Jing Zhang <zhangjing@microsoft.com>

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)

### Back port request
- [ ] 201911

We want to make sure switchover on both ToRs happens at the same time, to minimize the overhead and package loss time. 

#### How did you do it?

Force mux state switch to standby. 

Tested it on physical dual testbeds for 2000 switchovers, was able to minimize the gap from 67 ms to 1.3 ms.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,6 +8,7 @@ trigger:
     include:
       - "*"
 
+
 jobs:
 - job:
   displayName: "amd64"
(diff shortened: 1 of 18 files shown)


zjswhhh:
@yxieca - I used to test this change with the link prober intermediate state change together (in one build), but tried testing these two features separately this week. It turns out that the improvement from intermediate link prober states was very minor, we might need to reevaluate that design or proposal.

lguohan (approved):

yxieca (approved):

PR #11, opened 2021-12-09
Add TLV to the ICMP payload

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Add TLV support for the ICMP payload.

#### How did you do it?
This PR introduces a few changes to `IcmpPayload`
1. Change the `IcmpPayload` format as:
```
|cookie(4bytes)|version(4bytes)|uuid(8bytes)|seq(8bytes)|
```
2. Introduce `TlvCommand` and `TlvSentinel`
* `TlvCommand` is used for signaling peer ToR to switch status
```
|type(1bytes) = 0x1|length(2bytes) = 1|command(1bytes)|
```
* `TlvSentinel` is used for marking the end of the ICMP packet
```
|type(1bytes) = 0xff|length(2bytes) = 0|
```


lguohan:
in general, can we get better description for the pr?

lolyu:
> in general, can we get better description for the pr?

Updated, thanks!

src/link_prober/IcmpPayload.h
```diff
@@ -42,19 +42,76 @@ __BEGIN_DECLS
...
 {
+/**
+ *@enum TlvType
+ *
+ *@brief Supported TLV types
+ */
+enum TlvType: uint8_t {
+    TLV_COMMAND = 1,
+    TLV_SENTINEL = 0x0f,
+    TLV_COUNT
```
bingwang-ms, line 53: What's the `TLC_COUNT` used for?

src/link_prober/IcmpPayload.h
```diff
@@ -42,19 +42,76 @@ __BEGIN_DECLS
...
 namespace link_prober
 {
+/**
+ *@enum TlvType
+ *
+ *@brief Supported TLV types
+ */
+enum TlvType: uint8_t {
+    TLV_COMMAND = 1,
+    TLV_SENTINEL = 0x0f,
```
bingwang-ms, line 52: I suggest assigning `TLV_SENTINEL` to `0` since it's used as flag. And the type value larger than 0 is considered as common types.

src/link_prober/IcmpPayload.h
```diff
@@ -42,19 +42,76 @@ __BEGIN_DECLS
...
+    TLV_COUNT
+};
+
+#define TlvHead      \
+    uint8_t type;    \
+    uint16_t length
```
bingwang-ms, line 58: Why not define `TlvHead` as a new struct?

src/link_prober/IcmpPayload.h
```diff
@@ -130,6 +184,10 @@ struct IcmpPayload {
 static_assert(sizeof(IcmpPayload) % 2 == 0,
               "ICMP Payload size should be even sized, please add zero padding");
 
+static_assert(sizeof(TlvCommand) % 2 == 0, "TLV size should be even sized, please add zero padding");
```
bingwang-ms, line 187: Why `TlvSentinel` and `TlvCommand` must be even sized? We may combine several TLC struct in a ICMP payload.

src/link_prober/LinkProber.cpp
```diff
@@ -232,18 +232,18 @@ void LinkProber::handleUpdateEthernetFrame()
 //
 void LinkProber::handleSendSwitchCommand()
 {
-    IcmpPayload *icmpPayload = reinterpret_cast<IcmpPayload *> (
-        mTxBuffer.data() + sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr)
-    );
-    icmpPayload->command = htonl(static_cast<uint32_t> (Command::COMMAND_SWITCH_ACTIVE));
+    new (mTxBuffer.data() + sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload)) TlvCommand();
```
bingwang-ms, line 235: I suggest adding another variable like `icmp_payload_offset = sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload)` so that we don't have to do same calculation each time.

src/link_prober/LinkProber.cpp
```diff
@@ -232,18 +232,18 @@ void LinkProber::handleUpdateEthernetFrame()
...
+    computeChecksum(icmpHeader, sizeof(icmphdr) + sizeof(IcmpPayload) + sizeof(TlvCommand) + sizeof(TlvSentinel));
+    mTxPacketSize = sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload) + sizeof(TlvCommand) + sizeof(TlvSentinel);
 
     sendHeartbeat();
 
-    icmpPayload->command = htonl(static_cast<uint32_t> (Command::COMMAND_NONE));
-    computeChecksum(icmpHeader, sizeof(icmphdr) + sizeof(*icmpPayload));
+    new (mTxBuffer.data() + sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload)) TlvSentinel();
+    computeChecksum(icmpHeader, sizeof(icmphdr) + sizeof(IcmpPayload) + sizeof(TlvSentinel));
+    mTxPacketSize = sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload) + sizeof(TlvSentinel);
```
bingwang-ms, line 246: What's the purpose of these three lines?

src/link_prober/LinkProber.cpp
```diff
@@ -313,26 +308,42 @@ void LinkProber::handleRecv(
                 mMuxPortConfig.getPortName() %
                 mMuxPortConfig.getBladeIpv4Address().to_string()
             );
-            if (memcmp(icmpPayload->un.uuid.data, IcmpPayload::getGuidData(),
-                       IcmpPayload::getGuid().size()) == 0) {
+            bool isMatch = memcmp(icmpPayload->uuid, IcmpPayload::getGuidData(), sizeof(icmpPayload->uuid));
+            if (isMatch)
```
bingwang-ms, line 312: If icmpPayload->uuid == icmpPayload->uuid, isMatch will be zero.
Please double confirm if the logic is expected.

src/link_prober/LinkProber.cpp
```diff
@@ -313,26 +308,42 @@ void LinkProber::handleRecv(
...
-                        static_cast<void (LinkProberStateMachine::*) (SwitchActiveRequestEvent&)>
-                            (&LinkProberStateMachine::processEvent),
-                        &mLinkProberStateMachine,
-                        LinkProberStateMachine::getSwitchActiveRequestEvent()
-                    )));
+            }
+
+            uint8_t tlvStartOffset = sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload);
+            while (tlvStartOffset + sizeof(Tlv) < MUX_MAX_ICMP_BUFFER_SIZE)
+            {
```
bingwang-ms, line 322: I think the max length is supposed to be `bytesTransferred` rather than `MUX_MAX_ICMP_BUFFER_SIZE`

src/link_prober/LinkProber.cpp
```diff
@@ -313,26 +308,42 @@ void LinkProber::handleRecv(
...
+                                LinkProberStateMachine::getSwitchActiveRequestEvent()
+                            )));
+                        }
+                    }
+                    tlvStartOffset += sizeof(TlvCommand);
+                } else if (currentTlv->type == TlvType::TLV_SENTINEL) {
+                    // sentinel TLV, stop processing
+                    break;
+                } else {
+                    // unknown TLV type, ignore
```
bingwang-ms, line 343: For better compatibility, I suggest ignoring the unknown type and keep parsing.

src/link_prober/LinkProber.cpp
```diff
@@ -313,26 +308,42 @@ void LinkProber::handleRecv(
...
-                        &mLinkProberStateMachine,
-                        LinkProberStateMachine::getSwitchActiveRequestEvent()
-                    )));
+            }
+
+            uint8_t tlvStartOffset = sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload);
+            while (tlvStartOffset + sizeof(Tlv) < MUX_MAX_ICMP_BUFFER_SIZE)
+            {
+                Tlv *currentTlv = reinterpret_cast<Tlv *> (mTxBuffer.data() + tlvStartOffset);
+                if (currentTlv->type == TlvType::TLV_COMMAND) {
```
bingwang-ms, line 324: For better readaibility, two suggestions
1. Change the `if-else` to `switch-case`
2. Add a handler function for each TLV, such as `handle_tlv_command()`.

src/link_prober/LinkProber.cpp
```diff
@@ -551,10 +562,15 @@ void LinkProber::initializeSendBuffer()
...
+    icmphdr *icmpHeader = reinterpret_cast<icmphdr *> (mTxBuffer.data() + sizeof(ether_header) + sizeof(iphdr));
+    IcmpPayload *icmpPayload = new (mTxBuffer.data() + sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr)) IcmpPayload();
+    TlvSentinel *tlvSentinel = new (mTxBuffer.data() + sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload)) TlvSentinel();
+    mTxPacketSize = sizeof(ether_header) + sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload) + sizeof(TlvSentinel);
+
     ipHeader->ihl = sizeof(iphdr) >> 2;
     ipHeader->version = IPVERSION;
     ipHeader->tos = 0xb8;
-    ipHeader->tot_len = htons(sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload));
+    ipHeader->tot_len = htons(sizeof(iphdr) + sizeof(icmphdr) + sizeof(IcmpPayload) + sizeof(TlvSentinel));
```
bingwang-ms, line 573: Why the total len is fixed? We may have multi TLV structs inside.

src/link_prober/IcmpPayload.h
lolyu: removed

src/link_prober/IcmpPayload.h
lolyu: Uninitialized buffer has zeros inside, so I defined sentinel type to be `0xff`.

src/link_prober/IcmpPayload.h
lolyu: I use MACRO here because it will not introduce an extra layer

src/link_prober/IcmpPayload.h
lolyu: Now we assume all TLVs are even-sized so it is more easier to calculate the checksum

src/link_prober/IcmpPayload.cpp
```diff
@@ -38,6 +38,28 @@ boost::uuids::uuid IcmpPayload::mGuid;
 uint32_t IcmpPayload::mCookie = 0x47656d69;
 uint32_t IcmpPayload::mVersion = 0;
 
+//
+// ---> TlvCommand();
+//
+// struct TlvCommand constructor
+//
+TlvCommand::TlvCommand() :
+    type(tlvtype), length(htons(1)), command(static_cast<uint8_t> (Command::COMMAND_SWITCH_ACTIVE))
```
yxieca, line 47: needs to be packed.

I suggest adding a generic class with set_type(), set(length), set_value(), and the value field could be an union of {uint8_t value[]; command_t cmd; ...}.

src/link_prober/LinkProber.cpp
```diff
@@ -313,25 +338,41 @@ void LinkProber::handleRecv(
...
-                        static_cast<void (LinkProberStateMachine::*) (SwitchActiveRequestEvent&)>
-                            (&LinkProberStateMachine::processEvent),
-                        &mLinkProberStateMachine,
-                        LinkProberStateMachine::getSwitchActiveRequestEvent()
-                    )));
+            }
+
+            uint8_t tlvStartOffset = mPacketHeaderSize + sizeof(IcmpPayload);
+            bool stopProcessTlv = false;
+            while (tlvStartOffset + sizeof(Tlv) <= bytesTransferred) {
```
yxieca, line 352: Suggest adding member function of append_tlv and get_next_tlv to encapsulate the operations and add unit test to test these 2 methods generically.

src/link_prober/LinkProber.cpp
```diff
@@ -313,25 +338,41 @@ void LinkProber::handleRecv(
...
+                    default: {
+                        // unknonw TLV, try to skip
+                        if (currentTlvLength == 0) {
+                            stopProcessTlv = true;
+                            break;
+                        }
+                    }
+                    tlvStartOffset += sizeof(Tlv) + currentTlvLength;
+                }
+                if (stopProcessTlv) {
```
yxieca, line 374: add this to the loop condition directly?

src/link_prober/LinkProber.cpp
```diff
@@ -313,25 +338,41 @@ void LinkProber::handleRecv(
...
+                    break;
                 }
             }
```
yxieca, line 377: Add a debug event here if we didn't reach the end of the received bytes, but stopped by sentinel. state how many bytes missed.

src/link_prober/IcmpPayload.cpp
lolyu: For the tlv with variable size, will it hard for `union` to describe its contents?

src/link_prober/LinkProber.cpp
lolyu: updated, thanks!

src/link_prober/LinkProber.cpp
lolyu: added, thanks!

src/link_prober/LinkProber.cpp
lolyu: added methods:
* `findNextTlv`: find the next TLV from the offset
* `appendTlvCommand`: append TlvCommand to the end of tx buffer
* `appendTlvSentinel`: append TlvSentinel to the end of tx buffer

src/link_prober/LinkProber.cpp
```diff
@@ -232,18 +232,23 @@ void LinkProber::handleUpdateEthernetFrame()
...
-    computeChecksum(icmpHeader, sizeof(icmphdr) + sizeof(*icmpPayload));
+    computeChecksum(icmpHeader, sizeof(icmphdr) + totalPayloadSize);
+    ipHeader->tot_len = htons(sizeof(iphdr) + sizeof(icmphdr) + totalPayloadSize);
 
     sendHeartbeat();
 
-    icmpPayload->command = htonl(static_cast<uint32_t> (Command::COMMAND_NONE));
-    computeChecksum(icmpHeader, sizeof(icmphdr) + sizeof(*icmpPayload));
+    resetTxBufferTlv();
+    appendTlvSentinel();
```
zjswhhh, line 248: Do we only append TlvCommand when it's COMMAND_SWITCH_ACTIVE instead of having it in every heartbeat (COMMAND_NONE)? Shouldn't we have it in every heartbeat?

src/link_prober/LinkProber.cpp
```diff
@@ -591,4 +644,50 @@ void LinkProber::updateIcmpSequenceNo()
...
+//
+// Append TlvCommand to the end of txBuffer
+//
+size_t LinkProber::appendTlvCommand(Command commandType)
+{
+    assert(mTxPacketSize + sizeof(TlvCommand) <= MUX_MAX_ICMP_BUFFER_SIZE);
+    TlvCommand *tlvCommand = reinterpret_cast<TlvCommand *> (mTxBuffer.data() + mTxPacketSize);
+    tlvCommand->type = TlvCommand::tlvtype;
+    tlvCommand->length = htons(1);
+    tlvCommand->command = static_cast<uint8_t> (Command::COMMAND_SWITCH_ACTIVE);
```
zjswhhh, line 673: If it's always COMMAND_SWITCH_ACTIVE why do we need the function parameter?

src/link_prober/LinkProber.cpp
lolyu: Yes, currently the logic is to put command tlv only if there is need to send `COMMAND_SWITCH_ACTIVE`. Since now we remove the fixed command field from `IcmpPayload` and wrap it as a TLV, no `TlvCommand` is the same a `TlvCommand` with `COMMAND_NONE`.

src/link_prober/LinkProber.cpp
lolyu: the parameter `commandType` has a default value as `COMMAND_SWITCH_ACTIVE`, would it be better to leave it in the function signature if there is more command type in the future?

src/link_prober/IcmpPayload.cpp
lolyu: updated, please check

src/link_prober/LinkProber.cpp
```diff
@@ -313,36 +343,51 @@ void LinkProber::handleRecv(
...
-                        &mLinkProberStateMachine,
-                        LinkProberStateMachine::getSwitchActiveRequestEvent()
-                    )));
+            }
+
+            size_t nextTlvOffset = mTlvStartOffset;
+            size_t nextTlvSize = 0;
+            bool stopProcessTlv = false;
+            while ((nextTlvSize = findNextTlv(nextTlvOffset, bytesTransferred)) > 0 && !stopProcessTlv) {
+                Tlv *nextTlvPtr = reinterpret_cast<Tlv *> (mRxBuffer.data() + nextTlvOffset);
```
yxieca, line 361: Here we need to check if tlv offset + length is still with received data range.

lolyu: The check is in `findNextTlv`, if the tlv exceeds the bytes transferred, `nextTlvSize` will set to zero.

yxieca (approved):

PR #13, opened 2021-12-15
Linkmgrd subscribing State DB route event 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to make linkmgrd subscribe events from ROUTE_TABLE in State DB, and react accordingly:  
- If any of the two default route state appears to be 'na', linkmgrd should switch to standby. 
- If both are 'ok', there will be a mux state probing and what happens next depends on linkmgrd state machine and the probing response.  

### Type of change

- [ ] Bug fix
- [x] New feature

To make linkmgrd subscribe state DB route event, and handle the switchovers. 

### Documentation 

Related PR: https://github.com/Azure/sonic-swss/pull/2009


src/MuxManager.cpp
```diff
@@ -234,6 +234,33 @@ void MuxManager::processProbeMuxState(const std::string &portName, const std::st
...
+    MUXLOGINFO(boost::format("%s: default route state: %s") % address % routeState);
+
+    if (address.is_v4()) {
+        mIpv4DefaultRouteState = routeState;
+    } else {
+        mIpv6DefaultRouteState = routeState;
+    }
+
+    std::string nextState = "na";
+    if (mIpv4DefaultRouteState == "ok" && mIpv6DefaultRouteState == "ok") {
```
yxieca, line 253: @lguohan do we have concern if either IPv4 or IPv6 BGP is not configured? Can we assume we will have them both eventually?

zjswhhh: Updated per discussion in Gemini sync-up meeting.

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -815,6 +815,29 @@ void LinkManagerStateMachine::handleSwitchActiveRequestEvent()
...
+// 
+// ---> handleDefaultRouteStateNotification(const std::string &routeState);
+// 
+// handle default route state notification from routeorch
+//
+void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::string &routeState)
+{
+    MUXLOGWARNING(boost::format("%s: state db default route state: %s") % mMuxPortConfig.getPortName() % routeState);
+
+    if(mComponentInitState.test(MuxStateComponent)) {
```
yxieca, line 827: add space after if

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -815,6 +815,29 @@ void LinkManagerStateMachine::handleSwitchActiveRequestEvent()
...
+        if (ms(mCompositeState) == mux_state::MuxState::Label::Active &&
+            routeState == "na") {
```
yxieca, line 829: I suggest putting '{' on a new line, especially when if condition expands multiple lines. new line helps to identify the beginning of the if body.

src/MuxManager.cpp
```diff
@@ -234,6 +234,34 @@ void MuxManager::processProbeMuxState(const std::string &portName, const std::st
...
+    }
+
+    std::string nextState = "na";
+    // For now we only need IPv4 default route state, this logic can be updated in the future.  
+    if (mIpv4DefaultRouteState == "ok") {
+        nextState = "ok";
+    }
+
+    PortMapIterator portMapIterator = mPortMap.begin();
+    while(portMapIterator != mPortMap.end()) {
```
yxieca, line 259: space after while please.

src/link_manager/LinkManagerStateMachine.cpp
zjswhhh: Updated accordingly.

src/link_manager/LinkManagerStateMachine.cpp
zjswhhh: Updated accordingly.

src/MuxManager.cpp
zjswhhh: Updated accordingly.

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -815,6 +815,28 @@ void LinkManagerStateMachine::handleSwitchActiveRequestEvent()
...
+// ---> handleDefaultRouteStateNotification(const std::string &routeState);
+// 
+// handle default route state notification from routeorch
+//
+void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::string &routeState)
+{
+    MUXLOGWARNING(boost::format("%s: state db default route state: %s") % mMuxPortConfig.getPortName() % routeState);
+
+    if (mComponentInitState.test(MuxStateComponent)) {
+        if (ms(mCompositeState) == mux_state::MuxState::Label::Active && routeState == "na") {
```
prsunny, line 828: I'm thinking we should check for `!= Standby`. What if mux is in a transient state and we got this event? What do you think?

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -815,6 +815,28 @@ void LinkManagerStateMachine::handleSwitchActiveRequestEvent()
...
+            CompositeState nextState = mCompositeState;
+            enterLinkProberState(nextState, link_prober::LinkProberState::Wait);
+            switchMuxState(nextState, mux_state::MuxState::Label::Standby);
```
prsunny, line 831: How does it behave if both ToRs have default route as "na" ?

zjswhhh: Both ToRs will switch to standby, the one switches first will be "active" at the end. Do we expect both ToRs to be "standby" in this scenario?

src/link_manager/LinkManagerStateMachine.cpp
zjswhhh: Yes I think this is a good point. I will evaluate the behavior and update accordingly.

prsunny (approved):

zjswhhh:
boost::asio::ip::make_address is unable to parse "::/0" and "0.0.0.0/0", and considering the key will always be these two strings, I updated the code and removed make_address(). Changes have been tested on virtual dual testbeds and worked as expected. 

@prsunny could you review again? Thanks!

PR #17, opened 2022-01-20
[linkmgrd] linkmgrd subscribes MUX_CABLE_INFO table to handle peer OIR events

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

In the current Gemini linkmgrd design, when ICMP pinging consistently fails, a link prober unknown event won't trigger a switchover. In this scenario, if active side's cable is unplugged, there won't be a toggle to switch standby side to active. This PR is to handle this edge case.  

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

To make sure switchover happens when active side's cable is unplugged even when ICMP is not working. 

#### How did you do it?
Make linkmgrd listen to MUX_CABLE_INFO table, which stores peer's link status. This peer's link status will be probed and updated through I2C from MUX every 60s. 

Added Unit tests. Also tested this change on dual testbed, toggling happened as expected.

### DIFF
--- a/src/MuxPort.cpp
+++ b/src/MuxPort.cpp
@@ -250,4 +250,20 @@ void MuxPort::handleMuxConfig(const std::string &config)
     )));
 }
 
+// 
+// ---> handleDefaultRouteState(const std::string &routeState);
+// 
+// handles default route state notification
+//
+void MuxPort::handleDefaultRouteState(const std::string &routeState)
+{
+    MUXLOGDEBUG(boost::format("port: %s, state db default route state: %s") % mMuxPortConfig.getPortName() % routeState);
+
+    boost::asio::io_service &ioService = mStrand.context();
+    ioService.post(mStrand.wrap(boost::bind(
+        &link_manager::LinkManagerStateMachine::handleDefaultRouteStateNotification,
+        &mLinkManagerStateMachine,
+        routeState
+    )));
+}
 } /* namespace mux */
--- a/src/MuxPort.cpp
+++ b/src/MuxPort.cpp
@@ -110,6 +110,28 @@ void MuxPort::handleLinkState(const std::string &linkState)
 }
 
 //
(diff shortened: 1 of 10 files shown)


zjswhhh:
Hi @vdahiya12 - could you please review this PR when you get a chance? Thanks!

src/MuxPort.cpp
```diff
@@ -109,6 +109,28 @@ void MuxPort::handleLinkState(const std::string &linkState)
...
+//
+// ---> handlePeerLinkState(const std::string &linkState); 
+//
+// handle peer's link state updates
+//
+void MuxPort::handlePeerLinkState(const std::string &linkState)
+{
+    MUXLOGDEBUG(boost::format("port: %s, state db peer link state: %s") % mMuxPortConfig.getPortName() % linkState);
+
+    link_state::LinkState::Label label = link_state::LinkState::Label::Down;
```
vdahiya12, line 121: Are you assuming in all cases that except an up it will be down? In that case we might be doing redundant things in case of an "N/A" which could also be a valid state. 
For Link state down we should exclusively tie it to "down";

vdahiya12:
Lets also test the changes in the lab by bringing the link down on the peer.

src/MuxPort.cpp
zjswhhh: Agree. I rethink about it, in the case of "N/A", we should assume peer's link state is up, to avoid unnecessary actions as this table is not always up to date. 

Updated accordingly.

zjswhhh:
> Lets also test the changes in the lab by bringing the link down on the peer.

I tested on virtual testbeds by manually changing the link_status_peer value in state db. Do you mean the physical testbed or?

zjswhhh:
Conducted tests on virtual testbeds with y_cable simulator reporting peer link state as down, it worked as expected. @vdahiya12

vdahiya12 (approved):

PR #21, opened 2022-02-09
[lgtm] Enable lgtm

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Enable lgtm for `sonic-linkmgrd` repo

#### How did you do it?
Add `lgtm.yml` to enable the process


src/LinkMgrdMain.cpp
```diff
@@ -59,6 +59,7 @@ namespace {
 int main(int argc, const char* argv[])
 {
     int retValue = EXIT_SUCCESS;
+    int dummyVal = -1;
```
lguohan, line 62: why do we need this?

lolyu: I was trying to trigger the lgtm check with some code change. Will remove it soon.

PR #22, opened 2022-02-11
Handle xcvrd crashing scenarios

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

XCVRD on a ToR can crash in different scenarios, and currently the expected behaviors are: 
1. Crashing when toggling to Active
   1. Toggle triggered by ```sudo config mux standby``` on peer 
      Expected to succeed. Peer will toggle mux. No action is needed. 
   1. Toggle triggered by ```sudo config mux active```
      Expected to fail. Nothing can be done to help.    
   1. Toggle Triggered by LinkProber: Unknown
      Expected to fail. Nothing can be done to help.  
   1. Toggle Triggered by peer link status down (OIR)
      Expected to fail. Nothing can be done to help.  
   1. linkmgrd mismatching last updated xcvrd state
      Can’t be sure if the last updated state is accurate, if Mux is: 
      a. Pointing to local ToR
            Expected to succeed. 
      b. Pointing to peer ToR
             Expected to fail. Local ToR ends in probing mux state, tunnel won't be created.        
1. Crashing when toggling to Standby
   1. Toggle triggered by LinkState: Down
      Expected to success. Peer will toggle mux. No action is needed. 
   1. Toggle triggered by ```sudo config mux mode standby```
	  Expected to success. Peer will toggle mux. No action is needed. 
   1. Toggle triggered by ```sudo config mux mode active``` on peer
      Expected to fail. Local ToR ends in probing mux state, tunnel won't be created. 
   1. Toggle triggered by default route state change
      Expected to fail. 
   1. linkmgrd mismatching last updated xcvrd state
      Can’t be sure if the last updated state is accurate, if Mux is: 
      a. Pointing to local ToR
            Expected to fail. Local ToR ends in probing mux state, tunnel won't be created.
      b. Pointing to peer ToR
             Expected to succeed.        
1. Crashing when probing mux state
	Probing is trigger by state changes, or state inconsistency. To guarantee the tunnel is created, better to toggle local ToR to standby in this case. 
1. Crashing in normal scenarios
    No action is needed. 

There is nothing we can do to help Case 1.ii, 1.iii, and 1.iv. 

This PR is to make local ToR toggle to standby in failure case  1.v.b, 2.iii, 2.iv, 2.v.a and 3.  

This fix relies on ICMP heartbeat. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

To make dual ToR still functional in most XCVRD crashing cases. 

#### How did you do it?
Send switch to active command to peer through ICMP. Set mux cable table in app db so orchagent can establish the tunnel. 

Tested on virtual testbed with ICMP_responder on and pmon off, peer successfully toggled to active in the case mentioned above.

### DIFF
--- a/src/link_manager/LinkManagerStateMachine.cpp
+++ b/src/link_manager/LinkManagerStateMachine.cpp
@@ -61,7 +61,7 @@
 namespace link_manager
 {
 
-constexpr auto MAX_BACKOFF_FACTOR = 8;
+constexpr auto MAX_BACKOFF_FACTOR = 32;
 
 LinkManagerStateMachine::TransitionFunction
     LinkManagerStateMachine::mStateTransitionHandler[link_prober::LinkProberState::Label::Count]
--- a/src/link_manager/LinkManagerStateMachine.cpp
+++ b/src/link_manager/LinkManagerStateMachine.cpp
@@ -859,6 +859,7 @@ void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::str
     if (mComponentInitState.test(MuxStateComponent)) {
         if (ms(mCompositeState) != mux_state::MuxState::Label::Standby && routeState == "na") {
             mSendPeerSwitchCommandFnPtr();
+            // In case Mux is in wait state, switchMuxSate(standby) will be skipped. Setting mux state in app db to be standby so tunnel can be established.
             mMuxPortPtr->setMuxState(mux_state::MuxState::Label::Standby);
         } else {
             enterMuxWaitState(mCompositeState);
--- a/src/link_manager/LinkManagerStateMachine.cpp
+++ b/src/link_manager/LinkManagerStateMachine.cpp
@@ -528,6 +528,8 @@ void LinkManagerStateMachine::handleStateChange(MuxStateEvent &event, mux_state:
             handleMuxConfigNotification(mTargetMuxMode);
             mPendingMuxModeChange = false;
         }
+
+        mMuxWaitTimeoutCount = 0;
     }
 
     if (ms(mCompositeState) != mux_state::MuxState::Unknown) {
@@ -856,11 +858,8 @@ void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::str
 
     if (mComponentInitState.test(MuxStateComponent)) {
         if (ms(mCompositeState) != mux_state::MuxState::Label::Standby && routeState == "na") {
-            CompositeState nextState = mCompositeState;
-            enterLinkProberState(nextState, link_prober::LinkProberState::Wait);
-            switchMuxState(nextState, mux_state::MuxState::Label::Standby, true);
-            LOGWARNING_MUX_STATE_TRANSITION(mMuxPortConfig.getPortName(), mCompositeState, nextState);
-            mCompositeState = nextState;
+            mSendPeerSwitchCommandFnPtr();
+            mMuxPortPtr->setMuxState(mux_state::MuxState::Label::Standby);
         } else {
             enterMuxWaitState(mCompositeState);
         }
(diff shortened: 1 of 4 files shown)


src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -974,10 +974,17 @@ void LinkManagerStateMachine::startMuxWaitTimer(uint32_t factor)
 void LinkManagerStateMachine::handleMuxWaitTimeout(boost::system::error_code errorCode)
 {
     if (errorCode == boost::system::errc::success) {
+        mMuxWaitTimeoutCount++;
         if (mMuxStateMachine.getWaitStateCause() == mux_state::WaitState::WaitStateCause::SwssUpdate) {
             MUXLOGTIMEOUT(mMuxPortConfig.getPortName(), "orchagent timed out responding to linkmgrd", mCompositeState);
         } else if (mMuxStateMachine.getWaitStateCause() == mux_state::WaitState::WaitStateCause::DriverUpdate) {
             MUXLOGTIMEOUT(mMuxPortConfig.getPortName(), "xcvrd timed out responding to linkmgrd", mCompositeState);
+            // on the 3rd timeout, send switch active command to peer
+            if (mMuxWaitTimeoutCount == mMuxPortConfig.getNegativeStateChangeRetryCount()) {
```
yxieca, line 983: Should the condition be >=?

zjswhhh: > Should the condition be >=?

In Line 991 the mux wait timer will be started again, basically linkmgrd will keep restarting the timer when it times out, and keep printing log message. So we don't want to repeat this action in later timeouts.

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -856,11 +858,9 @@ void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::str
 
     if (mComponentInitState.test(MuxStateComponent)) {
         if (ms(mCompositeState) != mux_state::MuxState::Label::Standby && routeState == "na") {
-            CompositeState nextState = mCompositeState;
-            enterLinkProberState(nextState, link_prober::LinkProberState::Wait);
-            switchMuxState(nextState, mux_state::MuxState::Label::Standby, true);
-            LOGWARNING_MUX_STATE_TRANSITION(mMuxPortConfig.getPortName(), mCompositeState, nextState);
-            mCompositeState = nextState;
+            mSendPeerSwitchCommandFnPtr();
```
yxieca, line 861: I don't fully understand this chunk of diff. Here we don't know if xcvrd is healthy or not. The change assumed that xcvrd is not healthy?

zjswhhh: Even if xcvrd is healthy, it leaves no harm to do the switch this way. Send peer switch command includes the chunk. 

mMuxPortPtr->setMuxState() is to handle the case when mux is in wait state: the chunk will be skipped in this case. When it's not in wait state, it's no harm the set the same value twice.

yxieca (approved):

src/link_manager/LinkManagerStateMachine.cpp
lguohan, line 983: what is the getNegativeStateChangeRetryCount? will this create false positives?

zjswhhh: NegativeStateChangeRetryCount is a configurable threshold. It's 3 by default. Will increase the timeout limit.

zjswhhh: Updated MAX_BACKOFF_FACTOR to 32 in line 64. Now each timeout will be 9.6 seconds.

PR #24, opened 2022-02-14
[make] Enable make extra includes

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Enable passing extra includes to `make`

#### How did you do it?
Using the override directive to keep the variable `INCLUDES` in the Makefile to append to the one passed from the command line.

With command: `$ make all INCLUDES='-I"/sonic/src"'`
Check the extra includes is used by `g++`:
```
g++ -std=c++17 -D__FILENAME__="mux_state/ActiveState.cpp" -DBOOST_LOG_USE_NATIVE_SYSLOG -DBOOST_LOG_DYN_LINK -I"/sonic/src" -I".//src" -I"/usr/include/libnl3/" -O3 -Wall -c -fmessage-length=0 -fPIC -flto -MMD -MP -MF"src/mux_state/ActiveState.d" -MT"src/mux_state/ActiveState.o" -o "src/mux_state/ActiveState.o" "src/mux_state/ActiveState.cpp"
```


zjswhhh:
Why do we want to enable passing extra include?

yxieca (approved):

lolyu:
> Why do we want to enable passing extra include?

Because for lgtm, the `swss-common` will be not installed into system default path. So it will be nice to enable `make` to accept extra includes.

PR #27, opened 2022-02-19
Incrementing tolerance on mux state inconsistency 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to handle ```config mux mode standby/active``` failure on healthy ports caused by unexpected mismatching mux probe response.

What I firstly observed is a mux state probing right after a switchover can be mismatching with state db, and there was no additional switchover happened between the toggle and probe. Conducted an experiment and further proved that there is a time gap between xcvrd's toggle and the mux probe state change. 

This can trigger an undesirable switchback, and eventually appears to be a failure of ```config mux mode```. 

* Issue
  For example, expected state change sequence of switching to active is: 
    ```
    standby (Link Prober), standby (Local Mux State) 
    -> wait, wait
    -> wait, active (receive mux state notification, indicating toggle completes) 
    -> active, active (receive self's heartbeat)
    ```
  It's normal to receive previous heartbeat from peer during the process, so the sequence can also be:
    ```
    standby, standby 
    -> wait, wait 
    -> standby, wait (receive previous HB from peer before toggle completes) 
    -> standby, active (receive mux state notification, indicating toggle completes) 
    -> state inconsistency triggers mux state probing through I2C 
    -> mux probe response: active + receive self's heartbeat 
    -> active, active
    ```
  But since we have the time gap mentioned above, a probe response can unexpectedly be ```standby```, linkmgrd will trigger another switchover to standby. 

* Fix
Previous, no matter how linkmgrd enter ```standby, active``` or ```active, standby``` composite state, an immediate mux probe will be triggered. 
Now, when entering ```standby, active``` or ```active, standby``` composite state, check: 
  * If entering by link prober state change, do mux probe immediately. 
     Cause with link prober state change, toggle on cable definitely happen already. 
  * If entering by mux state notification, start a 300 ms timer. 
     When timer expires, if inconsistency still occurs, do mux probe. 

What caused the delay? We will need further investigation. This PR is a workaround for unblocking purpose. 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

To avoid toggle failure on healthy cables. 

#### How did you do it?
When receiving mux state change (from the switchover), hold for 300ms before mux state probing. 

Scripted 1000 switchovers on 24 ports. Can't reproduce the issue any more.

### DIFF
--- a/src/link_manager/LinkManagerStateMachine.cpp
+++ b/src/link_manager/LinkManagerStateMachine.cpp
@@ -942,11 +942,11 @@ void LinkManagerStateMachine::handleMuxProbeTimeout(boost::system::error_code er
     if (errorCode == boost::system::errc::success &&
         (ps(mCompositeState) == link_prober::LinkProberState::Label::Wait ||
          ms(mCompositeState) == mux_state::MuxState::Label::Unknown ||
-         ls(mCompositeState) == link_state::LinkState::Label::Down) ||
+         ls(mCompositeState) == link_state::LinkState::Label::Down ||
          (ps(mCompositeState) == link_prober::LinkProberState::Label::Standby &&
          ms(mCompositeState) == mux_state::MuxState::Label::Active) ||
         (ps(mCompositeState) == link_prober::LinkProberState::Label::Active &&
-         ms(mCompositeState) == mux_state::MuxState::Label::Standby)) {
+         ms(mCompositeState) == mux_state::MuxState::Label::Standby))) {
         CompositeState currState = mCompositeState;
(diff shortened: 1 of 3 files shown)


lguohan:
please add more description on what exactly the change is?

zjswhhh:
> please add more description on what exactly the change is?

Updated accordingly.

zjswhhh:
@yxieca - updated per discussion offline.

yxieca (approved):

PR #31, opened 2022-03-01
Collect port cable type to use corresponding state machine

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Let `MuxManager` collect per-port cable type info from `MUX_PORT` table, and init corresponding state machine.

#### How did you do it?
During the db interface thread method `handleSwssNotification`, call `getPortCableType` to collect per-port cable type info and store in `MuxManager`, so the creation of `MuxPort` could choose the state machine based on the cable type.

### DIFF
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -585,6 +585,47 @@ void DbInterface::getServerIpAddress(std::shared_ptr<swss::DBConnector> configDb
 }
 
 //
+// ---> processPortCableType(std::vector<swss::KeyOpFieldsValuesTuple> &entries);
+//
+// process port cable type and build a map of port name to cable type
+//
+void DbInterface::processPortCableType(std::vector<swss::KeyOpFieldsValuesTuple> &entries)
+{
+    for (auto &entry: entries) {
+        std::string portName = kfvKey(entry);
+        std::string operation = kfvOp(entry);
+        std::vector<swss::FieldValueTuple> fieldValues = kfvFieldsValues(entry);
+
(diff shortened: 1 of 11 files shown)


src/DbInterface.cpp
```diff
@@ -584,6 +584,47 @@ void DbInterface::getServerIpAddress(std::shared_ptr<swss::DBConnector> configDb
...
+        std::string operation = kfvOp(entry);
+        std::vector<swss::FieldValueTuple> fieldValues = kfvFieldsValues(entry);
+
+        std::string field = "cable_type";
+        std::vector<swss::FieldValueTuple>::const_iterator cit = std::find_if(
+            fieldValues.cbegin(),
+            fieldValues.cend(),
+            [&field] (const swss::FieldValueTuple &fv) {return fvField(fv) == field;}
+        );
+        std::string portCableType = (cit != fieldValues.cend() ? cit->second : "active-standby");
```
zjswhhh, line 605: So if ```cable_type``` is missing,  the default is ```active-standby```. It's accurate for now, but moving forward to active-active, should the default be ```active-standby```?

zjswhhh (approved):
Changes looks good to me.

src/DbInterface.cpp
lolyu: For the future `active-active`, if we can guarantee that the `cable_type` is always set for entries in `MUX_CABLE`, we are good here, right?

PR #37, opened 2022-03-08
[NONFUNCTIONAL] Add `LinkProberStateMachineBase`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Since the future `active-standby` mode will introduce different behavior for link prober state machine, define a base class `LinkProberStateMachine` to maintain common interfaces.

#### How did you do it?
1. Add `LinkProberStateMachineBase`
2. Let `LinkProberStateMachine` inherit from `LinkProberStateMachineBase`


zjswhhh (approved):
Looks good to me!

yxieca (commented):
Why did base and derived class live in the same header file? What is preventing them in different files?

I keep it in the same header file because they share most of the codes. If it will be better to keep it in different header files, will update then.

lolyu:
> Why did base and derived class live in the same header file? What is preventing them in different files?

I keep it in the same header file because they share most of the codes. If it will be better to keep it in different header files, will update then.

yxieca (approved):

PR #39, opened 2022-03-11
Add port cable type getter for `MuxManager`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

If a port is not in the `MUX_CABLE` table, `linkmgrd` still receives link up state, and calls `addOrUpdateMuxPortLinkState` to try to create a `MuxPort` object, in this case, the creation will fail due to the not-existed port cable type entry in the port cable type map.

#### How did you do it?
Add a getter `getMuxPortCableType` to `MuxManager`, if the port cable type map doesn't have an entry, set it with the default port cable mode `active-standby`

`linkmgrd` functions without error on a dualtor setup testbed.

### DIFF
--- a/src/MuxManager.cpp
+++ b/src/MuxManager.cpp
@@ -314,6 +314,20 @@ void MuxManager::addOrUpdateDefaultRouteState(bool is_v4, const std::string &rou
 }
 
 //
+// ---> getMuxPortCableType(const std::string &portName);
+//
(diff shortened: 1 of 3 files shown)


zjswhhh (approved):
Looks good to me!

PR #41, opened 2022-03-14
Add a command line option to store logs into a separate file

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Support store logs into a separate log file.

#### How did you do it?
Add an option `-e` or `--extra_log_file`, if present, stores the logs from `linkmgrd` into `/var/log/mux/linkmgrd.log`.

Run `linkmgrd` with the option on/off

### DIFF
--- a/src/common/MuxLogger.cpp
+++ b/src/common/MuxLogger.cpp
@@ -86,7 +86,8 @@ MuxLoggerPtr MuxLogger::getInstance()
 void MuxLogger::initialize(
     std::string &prog,
     std::string &path,
-    boost::log::trivial::severity_level level
+    boost::log::trivial::severity_level level,
+    bool extraLogFile
 )
 {
     namespace trivial = boost::log::trivial;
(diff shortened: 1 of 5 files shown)


src/common/MuxLogger.cpp
```diff
@@ -99,12 +100,13 @@ void MuxLogger::initialize(
...
     boost::log::init_from_settings(settings);
 
-//    boost::filesystem::remove(path);
-
-//    boost::log::add_file_log(
-//        keywords::file_name = path,
-//        keywords::format = "[%TimeStamp%] [%Severity%] %Message%"
-//    );
+    if (extraLogFile) {
+        boost::filesystem::remove(path);
```
zjswhhh, line 104: Do we want to add some protection on this `remove()`? A try/catch maybe?

lolyu: If the file doesn't exist, `remove` will return `false`, for those exceptions that `remove` might throw(rw issue, OS error), we don't have to catch them, right?

zjswhhh: I guess it's fine, I assume we are using this argument only for dev/debug purpose.

zjswhhh (approved):
Looks good to me!

PR #43, opened 2022-03-16
Decrease link probing interval after switchover to better determine the overhead of a toggle

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to get more accurate timestamp of when toggle completes on mux.  

The method is to decrease link probing interval to 10ms after a switchover is triggered, and write the timestamp of link prober state change to state db ```LINK_PROBE_STATS table```.

When switchover is over, revert the probing interval change. If switchover does not complete within 400ms, revert the change as well. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

To better determine the overhead of a toggle. 

#### How did you do it?
Decrease link probing interval after switchover is triggered. 

Tested cases below on dual testbed: 
1. switchover succeeds, icmp_respnder is on. 
2. switchover completes but icmp_responder is off. 

In both cases, link prober events are posted to state db as expected. Link probing interval is decreased and reverted as expected.


src/link_prober/LinkProber.cpp
```diff
@@ -748,4 +749,61 @@ void LinkProber::resetIcmpPacketCounts()
...
+        MUXLOGWARNING(boost::format("%s: link prober timeout on waiting for expected ICMP event after switchover is triggered ") % mMuxPortConfig.getPortName());
+    }
+}
+
+//
+// ---> getProbingInterval
+// 
+// get link prober interval
+//
+uint32_t LinkProber::getProbingInterval()
```
lolyu, line 803: Would it better to have this function as `inline`?

lolyu (approved):
LGTM

src/link_prober/LinkProber.cpp
zjswhhh: Updated accordingly.

PR #44, opened 2022-03-18
Shutdown ICMP heartbeats when default route state is missing and ToR is in `auto` mode

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to address the expected behavior on default route states: 
1.	If in “manual” mode, default route state change should NOT trigger mux state change. 
2.	If in “auto” mode, default route state change to “na”, should trigger switchover to “standby”.
3.	Avoid switchovers to “active” if default route state remains “na”. 

Sign-off: Jing Zhang zhangjing@microsoft.com

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

To handle default route state change. Make a healthy peer always takes over when self's missing default route state. 

#### How did you do it?
Shutdown heartbeats when default route state is missing and the mode is `auto`. 

Also make sure to restart the heartbeats when ToR enters manual mode or default route state is `ok`. 
Tested on dual testbeds.


lolyu (approved):

qiluo-msft:

PR #47, opened 2022-03-21
Add `LinkProberStateMachineActiveActive` 

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Add `LinkProberStateMachineActiveActive` for dualtor active-active mode.

#### How did you do it?
1. Improve `LinkProber` by adding event report member functions for both `active-active` and `active-standby` mode.
For `active-active` mode:
if `LinkProber` receives self ICMP echo reply, it will report `IcmpSelfEvent` event, if not, it will report `IcmpUnknown` event;
if `LinkProber` receives peer ICMP echo reply, it will report `IcmpPeerActiveEvent` event, if not, it will report `IcmpPeerUnknownEvent` event.
 2. Add `LinkProberStateMachineActiveActive` to work if the port is in `active-active` mode, and it has virtual peer event handler member function overrides to process the peer event and post the peer link prober event to `LinkManagerStateMachine::handlePeerStateChange`.
 3. Add unittests to cover the process for both `active-active` and `active-standby`.


src/link_prober/LinkProber.cpp
```diff
@@ -748,4 +772,62 @@ void LinkProber::resetIcmpPacketCounts()
...
+//
+// ---> reportHeartbeatReplyNotReceivedActiveActive
+//
+// report heartbeat reply not received to active-active mode link prober state machine
+//
+void LinkProber::reportHeartbeatReplyNotReceivedActiveActive()
+{
+    if (mTxSeqNo != mRxSelfSeqNo) {
+        mLinkProberStateMachinePtr->postLinkProberStateEvent(LinkProberStateMachineBase::getIcmpUnknownEvent());
+    }
```
zjswhhh, line 827: Are we planning to keep the pck loss counts for `active-active` as well? If so are we missing `mIcmpUnknownEventCount++;`  here?

src/link_prober/LinkProberStateMachineBase.cpp
```diff
@@ -100,10 +119,20 @@ void LinkProberStateMachineBase::processEvent(T &t)
...
-        postLinkManagerEvent(nextLinkProberState);
+    if (nextLinkProberState == nullptr) {
+        MUXLOGERROR(
+            boost::format(
+                "%s: link prober state %d could not handle event"
+            ) %
+            mMuxPortConfig.getPortName() %
+            currentLinkProberState->getStateLabel()
+        );
+    } else {
```
zjswhhh, line 130: Can you explain under what circumstances it will enter this block of code?

src/link_prober/LinkProber.cpp
lolyu: Yes, I want to leave the telemetry features to future PRs, this PR only addresses the basic event handling && state processing functionality.

src/link_prober/LinkProberStateMachineBase.cpp
lolyu: In `LinkProberState`, I've changed the following event handler method functions to virtual functions and added a default implementation that will return a `nullptr`.
```cpp
LinkProberState::handleEvent(IcmpPeerEvent &event)
LinkProberState::handleEvent(IcmpSelfEvent &event)
LinkProberState::handleEvent(IcmpUnknownEvent &event)
LinkProberState::handleEvent(IcmpPeerActiveEvent &event)
LinkProberState::handleEvent(IcmpPeerUnknownEvent &event)
```
Actually, I believe this code block will never hit the condition: `nextLinkProberState == nullptr`, because we always assume self state only process self probe event and peer state only process peer probe event. This condition only acts as a failover check.

zjswhhh (approved):
LGTM!

yxieca:
@lolyu  can you fix the merge conflict?

PR #48, opened 2022-03-23
Decrease link probing interval after switchover to better determine the overhead of a toggle #43

zjswhhh:
### Description of PR

Original commit & PR in master branch: 

c43cf7a Jing Zhang      Tue Mar 22 16:22:00 2022 -0700  Decrease link probing interval after switchover to better determine the overhead of a toggle (#43)

Summary:
Fixes # (issue)

This PR is to get more accurate timestamp of when toggle completes on mux.  

The method is to decrease link probing interval to 10ms after a switchover is triggered, and write the timestamp of link prober state change to state db ```LINK_PROBE_STATS table```.

When switchover is over, revert the probing interval change. If switchover does not complete within 400ms, revert the change as well. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

To better determine the overhead of a toggle. 

#### How did you do it?
Decrease link probing interval after switchover is triggered. 

Tested cases below on dual testbed: 
1. switchover succeeds, icmp_respnder is on. 
2. switchover completes but icmp_responder is off. 

In both cases, link prober events are posted to state db as expected. Link probing interval is decreased and reverted as expected.

### DIFF
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -46,7 +46,7 @@ constexpr auto DEFAULT_TIMEOUT_MSEC = 1000;
 std::vector<std::string> DbInterface::mMuxState = {"active", "standby", "unknown", "Error"};
 std::vector<std::string> DbInterface::mMuxLinkmgrState = {"uninitialized", "unhealthy", "healthy"};
 std::vector<std::string> DbInterface::mMuxMetrics = {"start", "end"};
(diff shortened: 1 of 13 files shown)


yxieca (approved):
@zjswhhh please update the PR message to state the reason for the change, you can copy it from the original PR.

PR #49, opened 2022-03-29
Disable the feature that decreases link probe interval for measuring switch overhead

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

Disable part of the feature introduced in #43. 

The link probing interval will NOT be decreased by default. Link prober state change events will still be posted in `LINK_PROBE_STATS|PORTNAME` in state db. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

We need to reconsider the design of this feature. 

To be more specific, this is a special case of decreasing probing interval, it's for measurement purposes only. We still want to trigger the toggle in 300ms when pack loss happens. The negative count should be 30 instead of 3 when interval is decreased to 10ms.

### DIFF
--- a/src/common/MuxPortConfig.h
+++ b/src/common/MuxPortConfig.h
@@ -292,6 +292,15 @@ public:
      */
     inline void setPortCableType(PortCableType portCableType) { mPortCableType = portCableType; };
 
+    /**
+     * @method ifEnableSwitchoverMeasurement
+     * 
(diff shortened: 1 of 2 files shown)


vdahiya12 (approved):

qiluo-msft:

zjswhhh:
PR submitted and merged already.

PR #54, opened 2022-04-01
Disable the feature that decreases link probe interval for measuring switch overhead #49

zjswhhh:
### Description of PR

Can't cleanly cherry pick the commit from master branch: 
34a68d1 disable switchover measuring based on link prober (#49)

Summary:
Fixes # (issue)

Disable part of the feature introduced in #43. 

The link probing interval will NOT be decreased by default. Link prober state change events will still be posted in `LINK_PROBE_STATS|PORTNAME` in state db. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

We need to reconsider the design of this feature. 

To be more specific, this is a special case of decreasing probing interval, it's for measurement purposes only. We still want to trigger the toggle in 300ms when pack loss happens. The negative count should be 30 instead of 3 when interval is decreased to 10ms.

### DIFF
--- a/src/common/MuxPortConfig.h
+++ b/src/common/MuxPortConfig.h
@@ -260,6 +260,15 @@ public:
     */
     inline uint32_t getDecreasedTimeoutIpv4_msec() const {return mMuxConfig.getDecreasedTimeoutIpv4_msec();};
 
+    /**
+     * @method ifEnableSwitchoverMeasurement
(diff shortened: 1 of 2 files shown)


yxieca (approved):
This is not really what asked but it gets the feature disabled for now.

PR #55, opened 2022-04-01
Keep incrementing sequence number when link prober is suspended and shutdown 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is fix the issue when both sides' link prober is suspended, LinkManager still appears to be healthy. 

When HB suspends, TX sequence number is not updated either, hence link prober won't post an `unknown` event. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Make sure link prober is reporting `unknown` event when it's suspended or shutdown on both sides. 

#### How did you do it?
Increment sequence number even when HB is not sent. 

Tested on dualToR testbed.

### DIFF
--- a/src/link_prober/LinkProber.cpp
+++ b/src/link_prober/LinkProber.cpp
@@ -299,9 +299,10 @@ void LinkProber::sendHeartbeat()
 {
     MUXLOGTRACE(mMuxPortConfig.getPortName());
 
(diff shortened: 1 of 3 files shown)


yxieca (approved):

qiluo-msft:

PR #56, opened 2022-04-01
Reset link prober state when default route is back 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to address the edge case that when both sides are missing default route, and standby side gets default route back. Expected behavior is standby side will switch to active. But if standby side is `wait` state when link prober is restarted, the switchover won't happen. 

sign-off: Jing Zhang [zhangjing@microsoft.com](mailto:zhangjing@microsoft.com)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

When both sides are missing default route AND standby side gets default route back, we want to ensure the standby side will take over and switch to active. 

#### How did you do it?
Re-initialize link prober state when default route goes from `na` to `ok`. 

Tested below scenarios: 
1. Active side missing default route. 
2. Both sides missing default route one after another.
3. Both sides missing default route, standby side get it back.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -861,6 +861,9 @@ void ActiveStandbyStateMachine::handleDefaultRouteStateNotification(const std::s
 {
     MUXLOGWARNING(boost::format("%s: state db default route state: %s") % mMuxPortConfig.getPortName() % routeState);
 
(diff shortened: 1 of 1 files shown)


yxieca (approved):

qiluo-msft:

zjswhhh:
Hi @qiluo-msft - can you try `cherry-pick` this PR again. I merged a PR into 202012 earlier, it should resolve the conflict.

PR #58, opened 2022-04-05
Initializing default route as 'ok' in LinkManagerStateMachine

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

`mDefaultRouteState` is a value cached in LinkManagerStateMachine. Initializing it as `na` might shutdown `LinkProber` unexpectedly at the initialing stages, as the order of processing db entries notification is not fixed.  

signoff: Jing Zhang zhangjing@microsoft.com

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Avoiding oscillation in initialization. 

Run dualtor_io tests.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.h
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.h
@@ -903,7 +903,7 @@ private:
 
     bool mContinuousLinkProberUnknownEvent = false; // When posting unknown_end event, we want to make sure the previous state is unknown.
 
-    std::string mDefaultRouteState = "na";
+    std::string mDefaultRouteState = "ok";
 
     std::bitset<ComponentCount> mComponentInitState = {0};
 };


yxieca:
Not sure either init value is right. There are 2 events in play here:

1. set the initial value.
2. get the default route state (this might change as BGP module progresses) and change from na to ok.

I am not sure initial value is an issue here?

zjswhhh:
> Not sure either init value is right. There are 2 events in play here:
> 
> 1. set the initial value.
> 2. get the default route state (this might change as BGP module progresses) and change from na to ok.
> 
> I am not sure initial value is an issue here?

You're correct about the 2 events. Default route state's initial value can determine if Link Prober is started by default. 

- initial value == "ok" --> Link prober starts by default, shuts down when changing to "na".  
- initial value == "na" --> Link prober has a chance to be shutdown in initialization stage. 
   Triggers to shutdown link prober: 1. mux mode change 2. default route change 
   Criteria to shutdown link prober: mode == auto && defaultRoute == na

   Imagining link manager handles config db entries (mux mode == auto) before state db entries (default route == ok) when initializing, it can cause unexpected oscillation.

zjswhhh:
Per offline discussion, closing this PR.

PR #60, opened 2022-04-08
Remove dh-systemd dependency for upgrading to bullseye

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

The dependency on `dh-systemd` is blocking upgrading docker-mux to `bullseye`. Removing it now to unblock. 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,6 @@ Section: devel
 Priority: optional
 Maintainer: Tamer Ahmed <tamer.ahmed@microsoft.com>
 Build-Depends: debhelper (>= 8.0.0),
-               dh-systemd
 Standards-Version: 3.9.3
 Homepage: https://github.com/Azure/sonic-linkmgrd
 XS-Go-Import-Path: github.com/Azure/sonic-linkmgrd


saiarcot895:
For the binary package sectionm can you replace the Depends list with `${shlibs:Depends}`? That'll automatically determine the dependencies. Otherwise, that list would need to get updated for Bullseye.

saiarcot895:
Also, FYI, the azure pipeline for building linkmgrd is on Buster. You may wish to update it to Bullseye at some point.

azure-pipelines.yml
```diff
@@ -32,27 +32,27 @@ jobs:
...
+          libboost-coroutine-dev \
+          libboost-date-time-dev \
+          libboost-fiber-dev \
+          libboost-filesystem-dev \
+          libboost-graph-parallel-dev \
+          libboost-log-dev \
+          libboost-regex-dev \
           googletest \
           libgtest-dev \
+          libgcc-10-dev \
```
saiarcot895, line 54: This shouldn't need to be explicitly specified; this gets automatically installed with the compiler.

zjswhhh: Hi Sai, thanks for the tip! 

I have addressed above comments.

saiarcot895 (approved):

PR #62, opened 2022-04-11
Avoid proactively switching to `active` if default route is missing 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to avoid proactively switch to `active` if default route is missing. 

By proactively switching, it means:
* switches triggered by peer HB missing
* switches triggered by peer link state down 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

ToR missing default route can't be `active healthy`. Toggling to active at this moment will only cause unwanted packet loss. 

#### How did you do it?
Avoid the switch. 

Passed all sonic-mgmt `dualtor_io` tests.

### DIFF
--- a/src/MuxPort.cpp
+++ b/src/MuxPort.cpp
@@ -263,7 +263,7 @@ void MuxPort::handleMuxConfig(const std::string &config)
 //
 void MuxPort::handleDefaultRouteState(const std::string &routeState)
 {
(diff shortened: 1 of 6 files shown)


qiluo-msft:

PR #64, opened 2022-04-13
Add `ActiveActiveStateMachine` implementation

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Add link manager state machine `ActiveActiveStateMachine` to work if the port is in `active-active` cable type.

#### How did you do it?
Add `ActiveActiveStateMachine` to cover the following scenarions:
* if the mux state is active, and link prober receives unknown event, linkmgrd will try to switch the mux to standby.
* if the mux state is standby, and link prober receives active event, linkmgrd will try to switch the mux to active.
* if the mux state is active, and link down event received, linkmgrd will try to switch the mux to standby.
* if the mux state is active, and mux config set to standby, linkmgrd will try to switch the mux to standby even with link prober active event.

Add unittests to cover the above scenarios.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -48,6 +48,7 @@ ActiveActiveStateMachine::ActiveActiveStateMachine(
            mux_state::MuxState::Label::Wait,
            link_state::LinkState::Label::Down}
       ),
+      mDeadlineTimer(strand.context()),
       mWaitTimer(strand.context()),
       mPeerWaitTimer(strand.context())
 {
@@ -239,6 +240,10 @@ void ActiveActiveStateMachine::handleProbeMuxStateNotification(mux_state::MuxSta
     MUXLOGWARNING(boost::format("%s: app db mux state: %s") % mMuxPortConfig.getPortName() % mMuxStateName[label]);
 
     mWaitTimer.cancel();
+    if (label == mux_state::MuxState::Label::Active || label == mux_state::MuxState::Label::Standby) {
+        mMuxProbeBackoffFactor = 1;
+        mDeadlineTimer.cancel();
+    }
 
     if (mComponentInitState.all()) {
         if (mMuxStateMachine.getWaitStateCause() != mux_state::WaitState::WaitStateCause::DriverUpdate) {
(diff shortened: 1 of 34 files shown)


src/link_manager/LinkManagerStateMachineActiveActive.cpp
```diff
@@ -0,0 +1,934 @@
...
+//
+// handles when xcvrd has timeout responding mux probe
+//
+void ActiveActiveStateMachine::handleMuxProbeTimeout(boost::system::error_code errorCode)
+{
+    MUXLOGDEBUG(mMuxPortConfig.getPortName());
+    if (errorCode == boost::system::errc::success) {
+        if (ms(mCompositeState) == mux_state::MuxState::Label::Unknown ||
+            ms(mCompositeState) == mux_state::MuxState::Label::Error ||
+            ms(mCompositeState) == mux_state::MuxState::Label::Wait) {
```
zjswhhh, line 846: Thought we got rid of `Wait` in active-active?

src/link_manager/LinkManagerStateMachineActiveActive.cpp
```diff
@@ -0,0 +1,934 @@
...
+            // suspend heartbeat to help peer ToR to toggle
+            mSuspendTxFnPtr(mMuxPortConfig.getLinkWaitTimeout_msec());
+        }
+        enterMuxState(nextState, label);
+        mMuxStateMachine.setWaitStateCause(mux_state::WaitState::WaitStateCause::SwssUpdate);
+        mMuxPortPtr->setMuxState(label);
+        mDeadlineTimer.cancel();
+        startMuxWaitTimer();
+    } else {
+        probeMuxState();
```
zjswhhh, line 714: if in non-auto mode, and probing back a mux state that is different than mux mode, will we switch to match mux mode?

zjswhhh: I guess the question is, is mux mode a "must-be" or a "preference"?

src/link_manager/LinkManagerStateMachineActiveActive.cpp
lolyu: Yes, `Wait` will only occur in the initial state `(Wait, Wait, Down)`

src/link_manager/LinkManagerStateMachineActiveActive.cpp
lolyu: Currently, the mux mode is more like a one-time-off setting, after setting the mux mode, `linkmgrd` now assumes the toggle must be a success, and won't react to any more state changes.

zjswhhh: Will it make more sense in `active-active` to keep mux state aligned with mux mode? But we can leave it as same as `active-standby` (one-time setting) for now.

zjswhhh (approved):
LGTM.

PR #67, opened 2022-04-20
[202012] Avoid proactively switching to active if default route is missing

zjswhhh:
### Description of PR

Can't cleanly `cherry-pick`  the commit below:
c4858a6 Jing Zhang      Thu Apr 14 11:27:55 2022 -0700  Avoid proactively switching to `active` if default route is missing  (#62)

Summary:
Fixes # (issue)

This PR is to avoid proactively switch to `active` if default route is missing. 

By proactively switching, it means:
* switches triggered by peer HB missing
* switches triggered by peer link state down 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

ToR missing default route can't be `active healthy`. Toggling to active at this moment will only cause unwanted packet loss. 

#### How did you do it?
Avoid the switch. 

Passed all sonic-mgmt `dualtor_io` tests.

### DIFF
--- a/src/link_manager/LinkManagerStateMachine.cpp
+++ b/src/link_manager/LinkManagerStateMachine.cpp
@@ -767,7 +767,9 @@ void LinkManagerStateMachine::handlePeerLinkStateNotification(const link_state::
     if(label == link_state::LinkState::Label::Down && ms(mCompositeState) == mux_state::MuxState::Standby) {
         CompositeState nextState = mCompositeState;
         enterLinkProberState(nextState, link_prober::LinkProberState::Wait);
-        switchMuxState(nextState, mux_state::MuxState::Label::Active);
+        if (mDefaultRouteState == DefaultRoute::OK) {
+            switchMuxState(nextState, mux_state::MuxState::Label::Active);
+        }
         LOGWARNING_MUX_STATE_TRANSITION(mMuxPortConfig.getPortName(), mCompositeState, nextState);
         mCompositeState = nextState;
     }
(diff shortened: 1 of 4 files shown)


src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -879,15 +881,15 @@ void LinkManagerStateMachine::handleSwitchActiveRequestEvent()
...
-// ---> handleDefaultRouteStateNotification(const std::string &routeState);
+// ---> handleDefaultRouteStateNotification(const DefaultRoute routeState);
 // 
 // handle default route state notification from routeorch
 //
-void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::string &routeState)
+void LinkManagerStateMachine::handleDefaultRouteStateNotification(const DefaultRoute routeState)
 {
-    MUXLOGWARNING(boost::format("%s: state db default route state: %s") % mMuxPortConfig.getPortName() % routeState);
+    MUXLOGDEBUG(mMuxPortConfig.getPortName());
```
yxieca, line 890: The new log message is almost useless?

yxieca, line 890: The new log message is almost useless?

src/link_manager/LinkManagerStateMachine.cpp
```diff
@@ -902,10 +904,10 @@ void LinkManagerStateMachine::handleDefaultRouteStateNotification(const std::str
 //
 void LinkManagerStateMachine::shutdownOrRestartLinkProberOnDefaultRoute()
 {
-    MUXLOGWARNING(boost::format("%s: default route state: %s") % mMuxPortConfig.getPortName() % mDefaultRouteState);
+    MUXLOGDEBUG(mMuxPortConfig.getPortName());
```
yxieca, line 907: What can you make out when you see this debug log?

yxieca, line 907: What can you make out when you see this debug log?

zjswhhh: > What can you make out when you see this debug log?

I used this log to confirm that there is check to shutdown or restart link prober when I updated mux mode or default route.

src/link_manager/LinkManagerStateMachine.cpp
zjswhhh: > The new log message is almost useless?

I used this kind of log to confirm notification was passed from `DbInterface` all the way to `LinkManagerStateMachine`, and handlers were executed.

lolyu (approved):

PR #73, opened 2022-05-12
Support switch between using wellknown mac or server mac addr 

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Add a new field `use_known_mac` to table `LINK_PROBER`:
* if `use_known_mac` is `enable`, linkmgrd will use known mac to probe(default behavior)
* if `use_known_mac` is `disable`, linkmgrd will use server mac address from netlink

#### How did you do it?
Add a new boolean attribute `mUseKnownMacActiveActive` to `MuxConfig` to represent this switch.
If `mUseKnownMacActiveActive` is `false`, the state machine will respond to server mac change.
else, the state machine will not respond to server mac change.

Pass unit test

### DIFF
--- a/src/common/MuxConfig.h
+++ b/src/common/MuxConfig.h
@@ -317,22 +317,22 @@ public:
     inline bool getIfEnableDefaultRouteFeature() {return mEnableDefaultRouteFeature;};
 
     /**
-     * @method getIfUseKnownMacActiveActive
+     * @method getIfUseWellKnownMacActiveActive
      * 
-     * @brief check if use known mac to probe for active-active ports
+     * @brief check if use well known mac to probe for active-active ports
      * 
-     * @return true to use known mac to probe for active-active ports
+     * @return true to use well known mac to probe for active-active ports
      */
-    inline bool getIfUseKnownMacActiveActive() { return mUseKnownMacActiveActive; };
+    inline bool getIfUseWellKnownMacActiveActive() { return mUseWellKnownMacActiveActive; };
 
     /**
-     * @method setUseKnownMacActiveActive
+     * @method setUseWellKnownMacActiveActive
      * 
-     * @brief setter to set if use known mac to probe for active-active ports
+     * @brief setter to set if use well known mac to probe for active-active ports
      * 
      * @return none
      */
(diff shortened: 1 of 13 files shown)


src/common/MuxConfig.h
```diff
@@ -316,6 +316,24 @@ class MuxConfig
...
     inline bool getIfEnableDefaultRouteFeature() {return mEnableDefaultRouteFeature;};
 
+    /**
+     * @method getUseKnownMacActiveActive
+     * 
+     * @brief check if use known mac to probe for active-active ports
+     * 
+     * @return true to use known mac to probe for active-active ports
+     */
+    inline bool getUseKnownMacActiveActive() { return mUseKnownMacActiveActive; };
```
zjswhhh, line 326: Better named as `getIfUseKnownMacActiveActive` as return value is a `bool`?

src/common/MuxConfig.h
```diff
@@ -316,6 +316,24 @@ class MuxConfig
...
+
+    /**
+     * @method setUseKnownMacActiveActive
+     * 
+     * @brief setter to set if use known mac to probe for active-active ports
+     * 
+     * @return none
+     */
+    inline void setUseKnownMacActiveActive(bool useKnownMacActiveActive) { mUseKnownMacActiveActive = useKnownMacActiveActive; };
```
zjswhhh, line 335: If config is updated to `false`, will previously updated server mac be consumed? No?

src/common/MuxConfig.h
lolyu: Updated

src/common/MuxConfig.h
lolyu: Now mux port config will cache the last mac-address update from Netlink, and use that if `useKnownMac` is set to `false`

yxieca:
Should be well-known mac address. And the well-known mac address is set to all 0s now?

lolyu:
> Should be well-known mac address. And the well-known mac address is set to all 0s now?

Sure, will update the name. The `mKnownMacAddress` has an initial value as all 0s, but it will be assigned with generated well-known mac for active-active ports.

lolyu:
@yxieca @zjswhhh , updated, please take a review again!

zjswhhh (approved):

PR #76, opened 2022-05-16
Add support for reconciliation after warm restart 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to add support for linkmgrd process reconciliation after warm restart. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

One step of warm reboot procedure for dual ToR is to config the switch into `manual` mode. Before warm reboot finalizer executes `config save`, we want to config the switch back into `auto` mode, so `config_db.json` will be consistent before and after the reboot. 

#### How did you do it?
1. When linkmgrd is initializing, get the systemwide warm reboot flag from `WARM_RESTART_ENABLE_TABLE`. If flag == true, start a reconciliation timer. 
2. Maintenance a mux port count based on `MUX_CABLE|PORTNAME` count. When one port completes reconciliation, if warm restart flag == true, config it back into `auto` mode, reduce reconciliation port count by 1.
3. If reconciliation timer expires or port count == 0, set state to `reconciled` in `WARM_RESTART_TABLE|linkmgrd`.

1. Unit tests
2. Tested on dual ToR testbed. Ports were `auto` mode after warm restart completed. Entry `WARM_RESTART_TABLE|linkmgrd` was added as expected.

### DIFF
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -715,6 +715,7 @@ void DbInterface::getServerIpAddress(std::shared_ptr<swss::DBConnector> configDb
     std::vector<swss::KeyOpFieldsValuesTuple> entries;
 
     configDbMuxCableTable.getContent(entries);
+    mMuxManagerPtr->updateWarmRestartReconciliationCount(entries.size());
     processServerIpAddress(entries);
 }
 
@@ -815,6 +816,52 @@ void DbInterface::getSoCIpAddress(std::shared_ptr<swss::DBConnector> configDbCon
 
     configDbMuxCableTable.getContent(entries);
     processSoCIpAddress(entries);
+}
+
+// ---> warmRestartReconciliation(const std::string &portName);
+//
+// port warm restart reconciliation procedure
+//
+void DbInterface::warmRestartReconciliation(const std::string &portName)
+{
+    MUXLOGDEBUG(portName);
+
+    setMuxMode(portName, "auto");
+    mMuxManagerPtr->updateWarmRestartReconciliationCount(-1);
+}
+
+//
+// ---> setMuxMode
+//
+// set config db mux mode
+//
+void DbInterface::setMuxMode(const std::string &portName, const std::string state)
+{
+    MUXLOGDEBUG(portName);
+
+    boost::asio::io_service &ioService = mStrand.context();
+    ioService.post(mStrand.wrap(boost::bind(
+        &DbInterface::handleSetMuxMode,
+        this,
+        portName,
+        state
+    )));
+}
+
+//
+// ---> handleSetMuxmode
+//
+// handle set mux mode
+//
+void DbInterface::handleSetMuxMode(const std::string &portName, const std::string state)
+{
+    MUXLOGWARNING(boost::format("%s: configuring mux mode to %s after warm restart") % portName % state);
+
+    std::shared_ptr<swss::DBConnector> configDbPtr = std::make_shared<swss::DBConnector> ("CONFIG_DB", 0);
+    std::shared_ptr<swss::Table> configDbMuxCableTablePtr = std::make_shared<swss::Table> (
+        configDbPtr.get(), CFG_MUX_CABLE_TABLE_NAME
+    );
+    configDbMuxCableTablePtr->hset(portName, "state", state);
 }
 
 //
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -826,8 +826,10 @@ void DbInterface::warmRestartReconciliation(const std::string &portName)
 {
     MUXLOGDEBUG(portName);
 
-    setMuxMode(portName, "auto");
-    mMuxManagerPtr->updateWarmRestartReconciliationCount(-1);
+    if (isWarmStart()) {
+        setMuxMode(portName, "auto");
+        mMuxManagerPtr->updateWarmRestartReconciliationCount(-1);
+    }
 }
 
 //
--- a/src/DbInterface.h
+++ b/src/DbInterface.h
@@ -34,6 +34,7 @@
 #include "swss/dbconnector.h"
 #include "swss/producerstatetable.h"
 #include "swss/subscriberstatetable.h"
+#include "swss/warm_restart.h"
 
 #include "link_manager/LinkManagerStateMachineActiveStandby.h"
 #include "mux_state/MuxState.h"
@@ -295,7 +296,34 @@ public:
      * 
      * @return none
      */
-    virtual void warmRestartReconciliation(const std::string &portName);
+    void warmRestartReconciliation(const std::string &portName);
+
+    /**
+     * @method isWarmStart
+     * 
+     * @brief is warm start or not
+     * 
+     * @return system flag for warm start context 
+     */
+    virtual bool isWarmStart(){return swss::WarmStart::isWarmStart();};
+
+    /**
+     * @method getWarmStartTimer
+     * 
+     * @brief get warm start time out in sec
+     * 
+     * @return timeout in sec 
+     */
+    virtual uint32_t getWarmStartTimer(){return swss::WarmStart::getWarmStartTimer("linkmgrd", "mux");};
+
+    /**
+     * @method setWarmStartStateReconciled
+     * 
+     * @brief set warm start state reconciled
+     * 
+     * @return none
+     */
+    virtual void setWarmStartStateReconciled(){swss::WarmStart::setWarmStartState("linkmgrd", swss::WarmStart::RECONCILED);};
 
 private:
     friend class test::MuxManagerTest;
--- a/src/DbInterface.h
+++ b/src/DbInterface.h
@@ -274,6 +274,29 @@ public:
     */
     void stopSwssNotificationPoll() {mPollSwssNotifcation = false;};
 
+    /**
+     * @method setMuxMode 
+     * 
+     * @brief set config db mux mode 
+     * 
+     * @param portName (in) MUX port name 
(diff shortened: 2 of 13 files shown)


src/MuxManager.cpp
```diff
@@ -470,4 +472,79 @@ void MuxManager::generateServerMac(uint16_t serverId, std::array<uint8_t, ETHER_
...
+    )));
+}
+
+// ---> handleWarmRestartReconciliationTimeout
+// 
+// handle warm restart reconciliationTimeout
+//
+void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::error_code errorCode)
+{
+    if (errorCode == boost::system::errc::success) {
```
vaibhavhd, line 543: Just making sure - the errorCode in case of timeout will be == `success`?

src/link_manager/LinkManagerStateMachineActiveStandby.cpp
```diff
@@ -426,6 +426,8 @@ void ActiveStandbyStateMachine::activateStateMachine()
         mStartProbingFnPtr();
 
         updateMuxLinkmgrState();
+
+        mMuxPortPtr->checkWarmRestart();
```
vaibhavhd, line 430: In this design won't the reconciliation timer start only when all ports are anyway reconciled? Shouldn't the timer start before ` mMuxPortPtr->checkWarmRestart();` is called (when the components are getting initialized)?

src/MuxManager.cpp
```diff
@@ -470,4 +472,79 @@ void MuxManager::generateServerMac(uint16_t serverId, std::array<uint8_t, ETHER_
...
+    } 
+}
+
+// ---> startWarmRestartReconciliationTimer
+//
+// start warm restart reconciliation timer
+// 
+void MuxManager::startWarmRestartReconciliationTimer()
+{
+    mReconciliationTimer.expires_from_now(boost::posix_time::seconds(
```
vaibhavhd, line 527: Is this a blocking call? If yes, then won't all ports just wait for timeout rather than checking for reconciliation while the timer is running?

vaibhavhd (commented):
Some comments posted to clear my doubts as I I am not well versed with this part of code.

Other comments:
Please define what `reconciliation` means in this context.
Will there be separate PR for warmboot_finalizer where it should also wait for `linkmgrd` to reconcile?

vaibhavhd:
Please address the PR merge conflicts.

src/common/MuxConfig.h
```diff
@@ -330,6 +339,7 @@ class MuxConfig
     uint32_t mDecreasedTimeoutIpv4_msec = 10;
 
     bool mEnableDefaultRouteFeature = false;
+    uint32_t mMuxReconciliationTimeout_sec = 90;
```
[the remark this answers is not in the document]
zjswhhh: I put 90s here only because of our meeting notes. In healthy system, reconciliation should finish within seconds.  

We can certainly reduce it to a small number. Is there anything I should take into consideration besides linkmgrd's own reconciliation overhead? Is 90s timeout a concern generally for warmreboot?

src/DbInterface.cpp
```diff
@@ -620,6 +620,7 @@ void DbInterface::getServerIpAddress(std::shared_ptr<swss::DBConnector> configDb
     std::vector<swss::KeyOpFieldsValuesTuple> entries;
 
     configDbMuxCableTable.getContent(entries);
+    mMuxManagerPtr->updateWarmRestartReconciliationCount(entries.size());
```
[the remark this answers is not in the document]
zjswhhh: At this stage linkmgrd won't be able to tell which one are unhealthy, it will probe all ports' mux state. The size of `entries` is the number of ports configured in `MUX_CABLE|PORTNAME` tables.

src/MuxManager.cpp
zjswhhh: Correct.

src/link_manager/LinkManagerStateMachineActiveStandby.cpp
zjswhhh: Yes `checkWarmRestart` is called after port is reconciled, but timer has been started before any port is reconciled. 

`checkWarmRestart` is to check if
* need to set port back to `auto` mode
* timer can be cancelled if this is the last port

src/MuxManager.cpp
zjswhhh: This is a non-blocking asynchronous wait.

src/DbInterface.h
```diff
@@ -395,6 +450,17 @@ class DbInterface
     */
     void getTorMacAddress(std::shared_ptr<swss::DBConnector> configDbConnector);
 
+    /**
+     * @method getWarmRestartFlag
+     * 
+     * @brief get flag to check if system is in warm reboot context is on
```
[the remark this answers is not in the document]
zjswhhh: Updated accordingly.

zjswhhh:
> Please address the PR merge conflicts.

Updated accordingly.

src/DbInterface.cpp
```diff
@@ -817,6 +818,54 @@ void DbInterface::getSoCIpAddress(std::shared_ptr<swss::DBConnector> configDbCon
...
+//
+// port warm restart reconciliation procedure
+//
+void DbInterface::warmRestartReconciliation(const std::string &portName)
+{
+    MUXLOGDEBUG(portName);
+
+    if (isWarmStart()) {
+        setMuxMode(portName, "auto");
+        mMuxManagerPtr->updateWarmRestartReconciliationCount(-1);
```
yxieca, line 831: Shouldn't we increment the count when in warm recovery?

yxieca (approved):

src/DbInterface.cpp
```diff
@@ -817,6 +818,54 @@ void DbInterface::getSoCIpAddress(std::shared_ptr<swss::DBConnector> configDbCon
...
+}
+
+//
+// ---> handleSetMuxmode
+//
+// handle set mux mode
+//
+void DbInterface::handleSetMuxMode(const std::string &portName, const std::string state)
+{
+    MUXLOGWARNING(boost::format("%s: configuring mux mode to %s after warm restart") % portName % state);
```
vaibhavhd, line 860: What's the reason for setting this as WARNING log? I am assuming that this is expected event, and hence INFO is more suited?

src/MuxManager.cpp
```diff
@@ -507,4 +514,65 @@ void MuxManager::generateServerMac(uint16_t serverId, std::array<uint8_t, ETHER_
...
+// 
+// handle warm restart reconciliationTimeout
+//
+void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::error_code errorCode)
+{
+    if (errorCode == boost::system::errc::success) {
+        MUXLOGWARNING("Reconciliation timed out after warm restart, set service to reconciled now.");
+    }
+
+    mDbInterfacePtr->setWarmStartStateReconciled();
```
vaibhavhd, line 575: In the case of timeout, some ports might be set as `auto` and some set as `manual`. Is that correct? If yes, is that expected? Should the timeout logic set the pending ports also as `auto`?

src/common/MuxConfig.h
```diff
@@ -387,6 +396,8 @@ class MuxConfig
     bool mEnableSwitchoverMeasurement = false;
     uint32_t mDecreasedTimeoutIpv4_msec = 10;
 
+    uint32_t mMuxReconciliationTimeout_sec = 10;
```
vaibhavhd, line 399: This may be a repeated question and may be answered already:

Should this timeout not be the function of number of ports on the device? For example, 28 ports vs 128 ports may make some considerable difference.

src/DbInterface.cpp
```diff
@@ -817,6 +818,54 @@ void DbInterface::getSoCIpAddress(std::shared_ptr<swss::DBConnector> configDbCon
...
+// ---> warmRestartReconciliation(const std::string &portName);
+//
+// port warm restart reconciliation procedure
+//
+void DbInterface::warmRestartReconciliation(const std::string &portName)
+{
+    MUXLOGDEBUG(portName);
+
+    if (isWarmStart()) {
+        setMuxMode(portName, "auto");
```
vaibhavhd, line 830: Note that per-port setting muxmode as `auto` can temporarily cause a subset of ports to be in auto, and others in manual mode.
I don't foresee any issues with that (do you?), but I am curious why you don't wait for all ports to reconcile (statemachine init done), and set all the ports together as `auto`? Also, that way setting them as auto would be consistent with setting them as manual (all at once).

vaibhavhd (commented):
Mostly LGTM, some non-blocking comments.

Will you raise a separate PR for sonic-utilities where FINALIZER would wait for linkmgrd to reconcile? Or is that not needed now since 10s timeout will trigger sooner than FINALIZER anyway?

src/common/MuxConfig.h
zjswhhh: I didn't see much time differences between 56 ports and 128 ports (both took around 6 seconds), think we should be good here.

vaibhavhd (approved):

zjswhhh:
> Mostly LGTM, some non-blocking comments.
> 
> Will you raise a separate PR for sonic-utilities where FINALIZER would wait for linkmgrd to reconcile? Or is that not needed now since 10s timeout will trigger sooner than FINALIZER anyway?

Thanks for the reminder! I think we need raise a separate PR for it, will do!

qiluo-msft:

PR #79, opened 2022-05-19
Check self's mux mode before switching peer to standby & add support for `detach` mode

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
Make sure self ToR is in `auto` mode before switching peer to `standby`. Also add support for `detach` mode. 

`detach` mode is equal to `auto` mode for SELF, but under this mode, ToR won't touch PEER's status. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -695,7 +695,8 @@ void ActiveActiveStateMachine::switchMuxState(
     mux_state::MuxState::Label label
 )
 {
-    if (mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Auto) {
+    if (mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Auto ||
+        mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Detached) {
         MUXLOGWARNING(
             boost::format("%s: Switching MUX state to '%s'") %
             mMuxPortConfig.getPortName() %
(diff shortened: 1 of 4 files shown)


src/link_manager/LinkManagerStateMachineActiveActive.cpp
```diff
@@ -722,14 +723,16 @@ void ActiveActiveStateMachine::switchMuxState(
...
-    startPeerMuxWaitTimer();
+    if (mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Auto) {
+        MUXLOGWARNING(
+            boost::format("%s: Switching peer MUX state to '%s'") %
+            mMuxPortConfig.getPortName() %
+            mMuxStateName[label]
+        );
+        enterPeerMuxState(label);
+        mMuxPortPtr->setPeerMuxState(label);
+        startPeerMuxWaitTimer();
```
vdahiya12, line 734: for the detach mode, could you please point , where is the actual logic to not query the peer ToR's gRPC API's ?

vdahiya12, line 734: for the detach mode, could you please point , where is the actual logic to not query the peer ToR's gRPC API's ?

zjswhhh: Currently querying SELF & PEER is using a same DB table entry. Per offline discussion, will make another PR to separate in the future. 

This PR only avoids setting PEER's state.

vdahiya12 (approved):

lolyu (approved):
LGTM

PR #98, opened 2022-07-25
TSA enhancement

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
This PR is an enhancement for TSA. When the CONFIG DB entry below is added, linkmgrd should be set to `standby`. 

```
    "BGP_DEVICE_GLOBAL": {
        "STATE": {
            "tsa_enabled": "false"
        }
    }
```

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

This PR is to enhance TSA logic, to handle the scenario when BGP_DEVICE_GLOBAL configuration is pushed directly without explicitly executing `TSA`.

#### How did you do it?
Make linkmgrd listen CONFIG DB `CFG_BGP_DEVICE_GLOBAL_TABLE_NAME` table, and config mux mode to `standby` once `tsa_enabled == true`.

1. Added unit tests.
2. Tested on dual testbeds with `redis-cli -n 4 hset "BGP_DEVICE_GLOBAL|STATE" tsa_enabled true`, was able to see logs from handlers and `show mux status` switch to `standby`.

### DIFF
--- a/src/MuxManager.cpp
+++ b/src/MuxManager.cpp
@@ -575,4 +575,21 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
     mDbInterfacePtr->setWarmStartStateReconciled();
 }
 
+//
+// ---> handleTsaEnableNotification
+//
+// handle TSA Enable Notification
+//
+void MuxManager::handleTsaEnableNotification(bool enable)
+{
+    if (enable) {
+        PortMapIterator portMapIterator = mPortMap.begin();
+
+        while (portMapIterator != mPortMap.end()) {
+            portMapIterator->second->handleTsaEnable();
+            portMapIterator ++;
+        }
+    }
+}
+
 } /* namespace mux */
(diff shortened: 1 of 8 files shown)


src/MuxManager.cpp
```diff
@@ -575,4 +575,21 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
...
+void MuxManager::handleTsaEnableNotification(bool enable)
+{
+    if (enable) {
+        PortMapIterator portMapIterator = mPortMap.begin();
+
+        while (portMapIterator != mPortMap.end()) {
+            portMapIterator->second->handleTsaEnable();
+            portMapIterator ++;
+        }
+    }
```
yxieca, line 592: When TSB is issued, we should come back to active, right?

zjswhhh: Add support to come back to `auto`. This way we will let the state machine to decide if it needs to toggle to active. And the behavior aligns with explicitly executing TSB. 

https://github.com/Azure/sonic-buildimage/pull/9726

yxieca (approved):

PR #104, opened 2022-08-08
Fix race condition caused by strand `wrap` method

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
This PR is to fix race condition introduced by `wrap` method of `boost::asio::io_service::strand`: 

> In the following case:
> 
> async_op_1(..., s.wrap(a));
> async_op_2(..., s.wrap(b));
>
> the completion of the first async operation will perform s.dispatch(a), and the second will perform s.dispatch(b), but the order in which those are performed is unspecified. That is, you cannot state whether one happens-before the other. Therefore, none of the above conditions are met and no ordering guarantee is made. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

To fix the bug that linkgmrd cleans mux metrics table __after__ setting mux state to orchagent. 

#### How did you do it?
1. Use ` boost::asio::post()` to replace `wrap`.
2. Add unit tests to cover this race condition scenario:
   * Remove override of `setMuxState`, `postMetricsEvent` in FakeDbinterface so the real implementation can be executed. 
   * Override  `setMuxState`, `postMetricsEvent` in FakeMuxPort as LinkManagerStateMachine tests use fake mux ports, so existing tests won't be broken. 
   * Initiate thread pool in MuxManager test to better mimic, and run 1000 times `setMuxState` & `postMetricsEvent` to reproduce.

Unit tests. Verified:
1. With `wrap`, issue was reproduced 14/1000 times. 
```
[  FAILED  ] 1 test, listed below:
[  FAILED  ] MuxManagerTest.DbInterfaceRaceConditionCheck
test/MuxManagerTest.cpp:1014: Failure
Value of: mDbInterfacePtr->mDbInterfaceRaceConditionCheckFailure
  Actual: true
Expected: false
...
```
2. With `boost::asio::post()`, issue was not reproducible.

### DIFF
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -70,12 +70,11 @@ void DbInterface::getMuxState(const std::string &portName)
 {
     MUXLOGDEBUG(portName);
 
-    boost::asio::io_service &ioService = mStrand.context();
-    ioService.post(mStrand.wrap(boost::bind(
+    boost::asio::post(mStrand, boost::bind(
         &DbInterface::handleGetMuxState,
         this,
         portName
-    )));
+    ));
 }
 
 //
@@ -87,13 +86,12 @@ void DbInterface::setMuxState(const std::string &portName, mux_state::MuxState::
 {
     MUXLOGDEBUG(boost::format("%s: setting mux to %s") % portName % mMuxState[label]);
 
-    boost::asio::io_service &ioService = mStrand.context();
-    ioService.post(mStrand.wrap(boost::bind(
+    boost::asio::post(mStrand, boost::bind(
         &DbInterface::handleSetMuxState,
         this,
         portName,
         label
-    )));
+    ));
 }
 
 //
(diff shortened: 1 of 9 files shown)


lolyu:
Could you elaborate on why posting mux metrics run before setting mux state is problematic?

zjswhhh:
> Could you elaborate on why posting mux metrics run after setting mux state is problematic?

When posting a "SwitchStart" event, linkmgrd will clean the whole mux metrics. And if orchagent reacts on "set mux state" before that, the `orch_switch_<active/standby>_start` timestamp will be removed.

Nikola observed this issue a lot in pilot cluster: 
```
24495 PORT        EVENT                          TIME
24496 ----------  -----------------------------  ---------------------------
 
24497 Ethernet52  linkmgrd_switch_standby_start  2022-Jul-25 09:53:29.214514
 
24498 Ethernet52  orch_switch_standby_end        2022-Jul-25 09:53:29.238199
 
24499 Ethernet52  xcvrd_switch_standby_start     2022-Jul-25 09:53:29.251985
 
24500 Ethernet52  xcvrd_switch_standby_end       2022-Jul-25 09:53:29.256608
 
24501 Ethernet52  linkmgrd_switch_standby_end    2022-Jul-25 09:53:29.259217
```

lolyu (approved):

yxieca:
@zjswhhh this change cannot be cherry picked cleanly to 202205, can you check?

PR #106, opened 2022-08-09
Backoff mux probing for server down scenario

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to backoff mux probing when both sides are unable to receive ICMP heartbeat. In this case, originally, state machine will do mux state probing each 300ms. It will be too much when more than one interface is having the same issue, and it will cause delay in switchovers. 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

Improve switchover performance, reduce packet loss. 

#### How did you do it?
* Increment backoff factor every time state machine enters 
`LinkProber: Wait/Unknown, MuxState: Active, LinkState: Up`
`LinkProber: Wait, MuxState: Standby, LinkState: Up` 
* Set mux probe timer based on the backoff factor. 
* Reset factor when exiting the state. 

Tested with 1000 switchovers. The performance improvement is comparable to fully disabling mux probing.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -792,6 +792,8 @@ void ActiveStandbyStateMachine::handleSuspendTimerExpiry()
         CompositeState currState = mCompositeState;
         enterMuxWaitState(mCompositeState);
         LOGINFO_MUX_STATE_TRANSITION(mMuxPortConfig.getPortName(), currState, mCompositeState);
+    } else {
+        mUnknownActiveUpBackoffFactor = 1;
     }
 }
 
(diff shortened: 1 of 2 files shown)


yxieca:
Is max backoff of 128 enough? 128 is 12.8 seconds?

zjswhhh:
> Is max backoff of 128 enough? 128 is 12.8 seconds?

The maximum interval will be 128* 3*100ms =  38.4 seconds.

yxieca (approved):

zjswhhh:
The change is in 202012 branch already, added the `included` tag. 
https://github.com/sonic-net/sonic-linkmgrd/commit/faceb93c1953321580771596be70765b131b03f7

PR #122, opened 2022-08-17
[active-active] Remove unnecessary mux wait timeout logs

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to remove repeating `handlePeerMuxWaitTimeout` and `handleMuxWaitTimeout` log message. 

When printing this log, linkmgrd doesn't do a retry or state change, so it's meaningless to repeat.  

When gRPC server is down, this message will become endless and cause unwanted syslog flood. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Remove unnecessary chatty log message. 

#### How did you do it?
Stop starting mux wait timer again after first timeout. 

Existing Unit tests.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -1012,11 +1012,6 @@ void ActiveActiveStateMachine::handleMuxWaitTimeout(boost::system::error_code er
         } else {
             MUXLOGTIMEOUT(mMuxPortConfig.getPortName(), "Unknown timeout reason!!!", mCompositeState);
         }
-        mMuxWaitBackoffFactor <<= 1;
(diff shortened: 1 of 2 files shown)


lolyu (commented):
How about retrying the timeout for a certain amout of times instead of removing all of them?

zjswhhh:
> How about retrying the timeout for a certain amout of times instead of removing all of them?

Checked with Vaibhav offline, the turnaround of gRPC requests should be 5~10ms, our first timeout is 300ms already. Besides, this timeout doesn't do anything except adding one more log, right?

zjswhhh:
@lolyu @yxieca - can you please help review on this PR? Thanks!

yxieca (approved):

lolyu (approved):

PR #136, opened 2022-09-22
[Active-Active] avoid being stuck in `unknown` after process init

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This is to address the issue that after linkmgrd process init, some ports can be stuck in `unknown` state. 

The cause was in the init stage, the first `unknown` event can only trigger a single mux probe instead of a "loop probe", and if this probe returns `unknown` again:   
---> mux state is unchanged (`unkown` -> `unkown`)
---> `MuxStateMachine` won't post an event to `LinkManagerStateMachine`
---> no state transition will be executed
---> no looping probe
---> stuck in unknown state. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Avoid being in unknown. 

#### How did you do it?
Enter `wait` state again if receive `unknown` in `Init` stage. 

Unit tests.


yxieca (approved):

lolyu (approved):
LGTM, it will be helpful to add some logs for this mux status change.

PR #171, opened 2023-02-17
[active-active] Toggle to standby if default route is missing

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

To avoid extra disruption after a BGP shutdown.
The reason is that, after a BGP shutdown, `linkmgrd` will stop heartbeats, and both local ToR and peer ToR will receive the heartbeat missing and toggle the local ToR to standby.
There is a chance that the peer ToR toggles local ToR to standby first, and the local ToR will probe the mux finding that itself mux state as `standby`, the local ToR will toggle itself back to active. When the local ToR receives the link prober unknown event, it will toggle itself into `standby` at last.

The event sequence:

bgp_down_before

#### How did you do it?
When `linkmgrd` receives default route missing event, make an extra toggle to `standby`, so local ToR's toggle to standby should be earlier than peer ToR's.
And the disruption total time could be reduced because `linkmgrd` makes the toggle immediately after receiving the default  route missing event instead of waiting for the link prober unknown event after shutdown the heartbeat sending.

The event sequence after:

bgp_down_after

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -1244,6 +1244,18 @@ void ActiveActiveStateMachine::handleDefaultRouteStateNotification(const Default
 
     mDefaultRouteState = routeState;
     shutdownOrRestartLinkProberOnDefaultRoute();
+
+    if (mComponentInitState.all() &&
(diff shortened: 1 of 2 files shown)


zjswhhh (approved):
LGTM

PR #183, opened 2023-03-14
[active-standby] avoid unnecessary mux state probe after configuring to `auto`

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

Issue was found in recent upgrade: ycable firmware would return `active` or `standby` to both sides. Hence one side limkgrd state will be `unhealthy` and block upgrade. 

This PR is to unblock upgrade by avoiding an immediate mux probe. Instead, wait for 300ms and if there is an inconsistency then, probe mux state. 

Cable firmware update or fix is out of the scope of this PR. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Unblock production upgrade.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -778,8 +778,7 @@ void ActiveStandbyStateMachine::handleMuxConfigNotification(const common::MuxPor
                 LOGWARNING_MUX_STATE_TRANSITION(mMuxPortConfig.getPortName(), mCompositeState, nextState);
                 mCompositeState = nextState;
             } else {
-                mMuxStateMachine.setWaitStateCause(mux_state::WaitState::WaitStateCause::DriverUpdate);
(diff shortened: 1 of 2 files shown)


yxieca:
@zjswhhh I am wondering if repeat the query when inconsistent is detected would be a better solution? Delaying query sounds like a work-around that could fail.

zjswhhh:
> @zjswhhh I am wondering if repeat the query when inconsistent is detected would be a better solution? Delaying query sounds like a work-around that could fail.

Hi Ying - when this issue happens, we do repeat the query because of the inconsistency (cable firmware keeps return false value which is inconsistent to heartbeat state). 

The true fix is to update cable firmware, but if this inconsistency doesn't resolve, we will have `unhealthy` label, and any upgrade  will just be blocked.

zjswhhh:
Also the query is not only delayed, it will **cancelled** if linkgrd state is `healthy` when timer expires.

yxieca (approved):

PR #184, opened 2023-03-16
[active-standby] Enforce switchover based on heartbeats when mux probe keeps failing

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

Fixing a common issue in production. Root cause is bad firmware failing to respond to mux toggle or mux query. 

For example, in below senario, we will have two `standby` ToRs in the end, both sides will have tunnel-route installed. 
``` 
UT0: active -> toggle standby -> toggles succeeds -> healthy 
LT0: standby > receive heartbeat event active due to peer's toggle -> probe mux state -> mux state unknown -> probe mux state -> repeating ...
```
The traffic loss in this scenario is evitable, as heartbeat behavior is very consistent and from that linkmgrd should be table to tell the current mux direction. 

The solution is when mux probing is the **ONLY** failing component, linkmgrd enforce a switchover based on heartbeat status. To ensure the heartbeat behaviour is consistent, we count how many times we enter the composite state **continuously**. 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Minimize traffic loss. 

Tested on dualtor DUT, with mux-simulator off. 
1. Toggle `active`
2. Toggle `standby`

Tunnel route was fixed into "one side active, one side standby" state.

### DIFF
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -47,7 +47,7 @@ std::vector<std::string> DbInterface::mMuxState = {"active", "standby", "unknown
 std::vector<std::string> DbInterface::mMuxLinkmgrState = {"uninitialized", "unhealthy", "healthy"};
 std::vector<std::string> DbInterface::mMuxMetrics = {"start", "end"};
 std::vector<std::string> DbInterface::mLinkProbeMetrics = {"link_prober_unknown_start", "link_prober_unknown_end", "link_prober_wait_start", "link_prober_active_start", "link_prober_standby_start"};
-std::vector<std::string> DbInterface::mActiveStandbySwitchCause = {"Peer_Heartbeat_Missing" , "Peer_Link_Down" , "Tlv_Switch_Active_Command" , "Link_Down" , "Transceiver_Daemon_Timeout" , "Matching_Hardware_State" , "Config_Mux_Mode"};
+std::vector<std::string> DbInterface::mActiveStandbySwitchCause = {"Peer_Heartbeat_Missing" , "Peer_Link_Down" , "Tlv_Switch_Active_Command" , "Link_Down" , "Transceiver_Daemon_Timeout" , "Matching_Hardware_State" , "Config_Mux_Mode", "Hardware_State_Unknown"};
 
 //
 // ---> DbInterface(mux::MuxManager *muxManager);
--- a/src/DbInterface.cpp
+++ b/src/DbInterface.cpp
@@ -234,7 +234,7 @@ void DbInterface::handlePostSwitchCause(
         boost::posix_time::ptime time
 )
 {
-    MUXLOGDEBUG(boost::format("%s: post switch cause %s") %
+    MUXLOGWARNING(boost::format("%s: post last switch cause %s") %
         portName %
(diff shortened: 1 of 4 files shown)


src/link_manager/LinkManagerStateMachineActiveStandby.cpp
```diff
@@ -1158,10 +1174,14 @@ void ActiveStandbyStateMachine::LinkProberStandbyMuxUnknownLinkUpTransitionFunct
 {
     MUXLOGINFO(mMuxPortConfig.getPortName());
 
+    mStandbyUnknownUpCount++;
+
     if ((ps(mCompositeState) != ps(nextState)) &&
         (ps(nextState) == link_prober::LinkProberState::Label::Active ||
          ps(nextState) == link_prober::LinkProberState::Label::Standby)) {
         enterMuxWaitState(nextState);
+    } else if (mStandbyUnknownUpCount == mMuxPortConfig.getNegativeStateChangeRetryCount()) {
```
yxieca, line 1183: Could this branch be missed if the code took the first if branch when condition is met?

zjswhhh: When it could, it should. Because there is a heartbeat state change here.

lolyu (approved):

src/DbInterface.cpp
```diff
@@ -234,7 +234,7 @@ void DbInterface::handlePostSwitchCause(
         boost::posix_time::ptime time
 )
 {
-    MUXLOGDEBUG(boost::format("%s: post switch cause %s") %
+    MUXLOGWARNING(boost::format("%s: post last switch cause %s") %
```
theasianpianist, line 237: Is this too verbose? What's the reason for the change?

src/link_manager/LinkManagerStateMachineActiveStandby.cpp
yxieca: Are you saying that the state change will reset the counter?

src/DbInterface.cpp
zjswhhh: > Is this too verbose? What's the reason for the change?

It will add one extra log for each switchover. 

Telemetry won't onboard switch cause data to Kusto anytime soon, so I wanted to have it in syslog for now.

src/link_manager/LinkManagerStateMachineActiveStandby.cpp
zjswhhh: It should have been reset. 

If it didn't get reset, the only case I can think of is user trigger a manual toggle in between, in that case, still we should skip the new branch.

xumia:
It did not list all the checks in job level, caused by Azp connection settings.
image

zjswhhh:
Hi @xumia - Thanks for checking! But now it seems to be hanging for overall status?

qiluo-msft:
Please add ADO number in PR description.

zjswhhh:
> Please add ADO number in PR description.

Done

PR #163, opened 2023-01-25
[active-active] update linkmgr health label definition

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to update the definition of `unhealthy` label for active-active interfaces, to be considered as "healthy":
* Link status must be up 
* Heartbeat response mux be received
* Server status should be in sync with ToR side status, **OR** server is unable to respond its status. 

sign-off: Jing Zhang zhngjing@microsoft.com 
### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

#### How did you verify/test it?
* Config interface to `standby` 
```
admin@**************:~$ show mux status Ethernet44
PORT        STATUS    SERVER_STATUS    HEALTH    HWSTATUS    LAST_SWITCHOVER_TIME
----------  --------  ---------------  --------  ----------  ---------------------------
Ethernet44  standby   standby          healthy   consistent  2023-Jan-25 01:11:13.228902
```

* Config interface to `auto` 
```
admin@**************:~$ show mux status Ethernet44
PORT        STATUS    SERVER_STATUS    HEALTH    HWSTATUS    LAST_SWITCHOVER_TIME
----------  --------  ---------------  --------  ----------  ---------------------------
Ethernet44  active    active           healthy   consistent  2023-Jan-25 01:14:26.909006
```

* Server status unknown
```
admin@**************:~$ show mux status Ethernet44
PORT        STATUS    SERVER_STATUS    HEALTH    HWSTATUS    LAST_SWITCHOVER_TIME
----------  --------  ---------------  --------  ----------  ---------------------------
Ethernet44  active    unknown          healthy   consistent  2023-Jan-25 01:14:26.909006

admin@**************:~$ show mux status Ethernet44
PORT        STATUS    SERVER_STATUS    HEALTH    HWSTATUS    LAST_SWITCHOVER_TIME
----------  --------  ---------------  --------  ----------  ---------------------------
Ethernet44  standby   unknown          healthy   consistent  2023-Jan-25 01:15:03.012912
```

* Status mismatch 
```
admin@**************:~$ show mux status Ethernet44
PORT        STATUS    SERVER_STATUS    HEALTH     HWSTATUS    LAST_SWITCHOVER_TIME
----------  --------  ---------------  ---------  ----------  ---------------------------
Ethernet44  active    standby          unhealthy  consistent  2023-Jan-25 01:12:25.175324
```

* shutdown interface 
```
[...]

PR #188, opened 2023-03-17
Ignore mux probe unknown for health calculation

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

If the mux is healthy(`active`, `active`, `up`), mux probe `unknown` will transit the mux into `unhealthy`.
Based on the mux health definition from PR: https://github.com/sonic-net/sonic-linkmgrd/pull/163, if the mux server is not responsive, the mux should remain healthy in such a case.

#### How did you do it?
1. if the mux port is healthy and the last mux notification is `unknown` from mux probe, ignore it in the health calculation.
2. add the healthy check in the periodical probe test case
3. add three more test case to cover more scenarios:
    * mux active, periodical probe unknown
    * mux standby, periodical probe active
    * mux standby, periodical probe unknown

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -176,6 +176,8 @@ void ActiveActiveStateMachine::handleMuxStateNotification(mux_state::MuxState::L
     MUXLOGWARNING(boost::format("%s: state db mux state: %s") % mMuxPortConfig.getPortName() % mMuxStateName[label]);
 
     mWaitTimer.cancel();
+
+    mLastMuxNotificationType = LastMuxNotificationType::MuxNotificationFromToggle;
     mLastMuxStateNotification = label;
 
     if (mComponentInitState.all()) {
@@ -265,6 +267,9 @@ void ActiveActiveStateMachine::handleMuxConfigNotification(const common::MuxPort
 void ActiveActiveStateMachine::handleProbeMuxStateNotification(mux_state::MuxState::Label label)
 {
     MUXLOGINFO(boost::format("%s: app db mux state: %s") % mMuxPortConfig.getPortName() % mMuxStateName[label]);
+
+    mLastMuxNotificationType = LastMuxNotificationType::MuxNotificationFromProbe;
+    mLastMuxProbeNotification = label;
 
     mWaitTimer.cancel();
     if (label == mux_state::MuxState::Label::Active || label == mux_state::MuxState::Label::Standby) {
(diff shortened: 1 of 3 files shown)


src/link_manager/LinkManagerStateMachineActiveActive.cpp
```diff
@@ -176,6 +176,8 @@ void ActiveActiveStateMachine::handleMuxStateNotification(mux_state::MuxState::L
     MUXLOGWARNING(boost::format("%s: state db mux state: %s") % mMuxPortConfig.getPortName() % mMuxStateName[label]);
 
     mWaitTimer.cancel();
+
+    mLastMuxNotificationType = LastMuxNotificationType::MuxNotificationFromToggle;
```
zjswhhh, line 180: If the mux notification is not posted to mux state machine (`postMuxStateEvent(label)`), should we still update `mLastMuxNotificationType`?

src/link_manager/LinkManagerStateMachineActiveActive.cpp
```diff
@@ -1069,7 +1074,10 @@ void ActiveActiveStateMachine::updateMuxLinkmgrState()
     Label label = Label::Unhealthy;
     if (ls(mCompositeState) == link_state::LinkState::Label::Up &&
         ps(mCompositeState) == link_prober::LinkProberState::Label::Active &&
-        (ms(mCompositeState) == mLastMuxStateNotification || mLastMuxStateNotification == mux_state::MuxState::Label::Unknown) &&
+        (ms(mCompositeState) == mLastMuxStateNotification ||
+         mLastMuxStateNotification == mux_state::MuxState::Label::Unknown ||
```
zjswhhh, line 1078: Why can't we only check last probe state? Say if we only check last probe state: 
Case 1: Last toggle: active, last probe unknown -> healthy
Case 2: Last toggle: unknown, last probe active -> toggle should be triggered -> last toggle: active, last probe: active.

src/link_manager/LinkManagerStateMachineActiveActive.cpp
lolyu: This is to align with the update of `mLastMuxStateNotification`

src/link_manager/LinkManagerStateMachineActiveActive.cpp
lolyu: For case 2, if the latest toggle has mux `unknown` returned, and the periodical mux probe returns `active` later, this will not trigger an extra toggle, this will be fixed in my next PR.
And we could not only check the last probe state for now because case 2 is not covered. Let's postpone this check change for a while.

zjswhhh (approved):

lolyu:
17593800

PR #191, opened 2023-03-23
[active-active] Add mux server state correction mechanism

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

There is no uniform mux server state correction for the state machine in
case if the mux toggle fails at first(mux server state not expected).
Let's add an uniform mux server state correction that is compatible with
previous codes.

The basic idea is that, when the mux probe is able to return meaningful
mux states(`active` or `standby`), let's add a check to see if the last toggle
succeeded or not. So let's define our mux toggle decision process as the
following:

```
If mux config is active or standby:
    if last toggle success:
        NOOP
    else:
        toggle based on the mux config
else:
    if mux status not matches heartbeats:
        toggle based on heartbeats
    else:
        if last toggle success:
            NOOP
        else:
            toggle based on the heartbeats
```

As we only consider mux probe `active` or `standby`, so only four
transition functions needs to be improved:

LinkProberActiveMuxActiveLinkUpTransitionFunction
LinkProberActiveMuxStandbyLinkUpTransitionFunction
LinkProberUnknownMuxActiveLinkUpTransitionFunction
LinkProberUnknownMuxStandbyLinkUpTransitionFunction

NOTE: mux probe `unknown` is not covered as the primary goal is to
guarantee a toggle success. Mux server down after toggle is not a
sufficient reason for us to update the mux server status and break the
healthy state.

1. Add UT.
2. testbed verification
* with heartbeats, mux active -> config reload -> mux active consistently
* without heartbeats, mux standby -> config reload -> mux standby consistently
* with heartbeats, mux active -> mux config standby -> config save -> config reload -> mux standby consistently
* without hearbeats, mux standby -> mux config active -> config save -> config reload -> mux active consistently

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -749,9 +749,9 @@ void ActiveActiveStateMachine::LinkProberActiveMuxActiveLinkUpTransitionFunction
 {
     MUXLOGINFO(mMuxPortConfig.getPortName());
     if (mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Standby) {
-        switchMuxState(nextState, mux_state::MuxState::Label::Standby, true);
-    } else if (ms(mCompositeState) != mux_state::MuxState::Active) {
-        switchMuxState(nextState, mux_state::MuxState::Label::Active);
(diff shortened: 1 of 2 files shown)


zjswhhh (approved):
LGTM

PR #197, opened 2023-04-06
[202012][active-standby] Enforce switchover based on heartbeats when mux probe keeps failing #184

zjswhhh:
Cherry-pick conflict for #184.

Summary:
Fixes # (issue)

Fixing a common issue in production. Root cause is bad firmware failing to respond to mux toggle or mux query.

For example, in below senario, we will have two `standby` ToRs in the end, both sides will have tunnel-route installed.
```
UT0: active -> toggle standby -> toggles succeeds -> healthy
LT0: standby > receive heartbeat event active due to peer's toggle -> probe mux state -> mux state unknown -> probe mux state -> repeating ...
```
The traffic loss in this scenario is evitable, as heartbeat behavior is very consistent and from that linkmgrd should be table to tell the current mux direction.

The solution is when mux probing is the **ONLY** failing component, linkmgrd enforce a switchover based on heartbeat status. To ensure the heartbeat behaviour is consistent, we count how many times we enter the composite state **continuously**.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -75,24 +75,6 @@ jobs:
     inputs:
       source: specific
       project: build
-      pipeline: Azure.sonic-buildimage.common_libs
-      runVersion: 'latestFromBranch'
-      runBranch: 'refs/heads/$(sourceBranch)'
-      path: $(Build.ArtifactStagingDirectory)/download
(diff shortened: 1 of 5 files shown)


lolyu (approved):
Could you pls format your PR description?

PR #199, opened 2023-04-10
Enable debug symbols

### LINKED ISSUE #196: [linkmgrd] No debug symbols included in the `docker-mux-dbg` `linkmgrd` binary
As the subject.
* `4a470f7851a3 ` is `docker-mux-dbg` container id.
```
$ docker exec -it 4a470f7851a3 bash
root@4a470f7851a3:/# md5sum /usr/sbin/linkmgrd
82456cb93e52686682eae60844a3ea9d  /usr/sbin/linkmgrd
root@4a470f7851a3:/#
exit
$ docker exec -it mux bash
root@svcstr-7050-acs-1:/# md5sum /usr/sbin/linkmgrd
82456cb93e52686682eae60844a3ea9d  /usr/sbin/linkmgrd
```

lolyu:
### Description of PR

Summary:
Fixes #196 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Enable `docker-mux-dbg` to have correct debug symbol files.

#### How did you do it?
As Debian passes debug options via environment variable `CXXFLAGS` or `CFLAGS` during package build, so let's add `-g` to the `CPP_FLAGS` if `CXXFLAGS` contains `-g`. So the final `linkmgrd` binary could have debug symbols.

use `sonic-buildimage` to build target `target/debs/bullseye/sonic-linkmgrd-dbgsym_1.0.0-1_amd64.deb`, and verify the symbol files:
```
$ ls -lah usr/lib/debug/.build-id/48/619bee180a75bafa2b1ee64a5a26c8c00ad0fe.debug
-rw-r--r-- 1 lolv lolv 11M Oct 26  2020 usr/lib/debug/.build-id/48/619bee180a75bafa2b1ee64a5a26c8c00ad0fe.debug
$ ls -lah usr/sbin/linkmgrd
-rwxr-xr-x 1 lolv lolv 998K Oct 26  2020 usr/sbin/linkmgrd
```
```
(gdb) exec-file usr/sbin/linkmgrd
(gdb) symbol-file usr/lib/debug/.build-id/7a/19e9cd44f81ddb0bbc3eab757319675ae606de.debug
Reading symbols from usr/lib/debug/.build-id/7a/19e9cd44f81ddb0bbc3eab757319675ae606de.debug...
(gdb) list main
warning: Source file is more recent than executable.
65
66          //
67          // Constants for command line argument strings:
68          //
69          boost::log::trivial::severity_level level;
70          bool extraLogFile = false;
71          bool measureSwitchover = false;
72          bool defaultRoute = false;
73
74          program_options::options_description description("linkmgrd options");
quit)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ JOBS := $(subst -j,,$(JOB_FLAG))
 
 # keep debug option only
 CPP_FLAGS := $(if $(findstring -g,$(CXXFLAGS)), -g)
-LD_FLAGS := $(if $(findstring -g,$(CXXFLAGS)), -g)
 
 release-targets: CPP_FLAGS := $(CPP_FLAGS) -O3 -Wall -c -fmessage-length=0 -fPIC -flto
 test-targets: CPP_FLAGS := $(CPP_FLAGS) -O0 -Wall -c -fmessage-length=0 -fPIC $(GCOV_FLAGS)
@@ -70,7 +69,7 @@ all: sonic-linkmgrd
 release-targets: $(OBJS) $(USER_OBJS) $(OBJS_LINKMGRD)
 	@echo 'Building target: $@'
 	@echo 'Invoking: GCC C++ Linker'
-	$(CXX) -pthread -o "$(LINKMGRD_TARGET)" $(OBJS) $(OBJS_LINKMGRD) $(USER_OBJS) $(LIBS) $(LD_FLAGS)
+	$(CXX) -pthread -o "$(LINKMGRD_TARGET)" $(OBJS) $(OBJS_LINKMGRD) $(USER_OBJS) $(LIBS)
 	@echo 'Finished building target: $@'
 	@echo ' '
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,8 +17,12 @@ MAKE_PID := $(shell echo $$PPID)
 JOB_FLAG := $(filter -j%, $(subst -j ,-j,$(shell ps T | grep "^\s*$(MAKE_PID).*$(MAKE)")))
 JOBS := $(subst -j,,$(JOB_FLAG))
 
-release-targets: CPP_FLAGS := -O3 -Wall -c -fmessage-length=0 -fPIC -flto
-test-targets: CPP_FLAGS = -O0 -Wall -c -fmessage-length=0 -fPIC $(GCOV_FLAGS)
+# keep debug option only
+CPP_FLAGS := $(if $(findstring -g,$(CXXFLAGS)), -g)
```
saiarcot895, line 21: Why only the `-g` flag, and not all flags?

Makefile
```diff
@@ -17,8 +17,12 @@ MAKE_PID := $(shell echo $$PPID)
...
+LD_FLAGS := $(if $(findstring -g,$(CXXFLAGS)), -g)
```
saiarcot895, line 22: `CXXFLAGS` shouldn't be used for `LD_FLAGS`; the two are for different purposes.

Makefile
lolyu: Debian passes `CXXFLAGS` and `CFLAGS` as `-g -O2 -ffile-prefix-map=/sonic/src/linkmgrd=. -fstack-protector-strong -Wformat -Werror=format-security`, keeping the debug flag only here is to ensure that other flags hardcoded by `linkmgrd` will not be overridden, for instance, `linkmgrd` uses `-O3` for the release target.

Makefile
lolyu: Removed, thanks!

Makefile
saiarcot895: For `-O2` vs `-O3`, I believe the last specified flag takes effect, so if `-O3` is specified after `-O2`, then `-O3` gets used.

lolyu: Yes, keeping both optimization flag in the compile command is a little bit confusing somehow, as we only need `-g` in the flags, is it better if we only keep `-g` here?

saiarcot895: Technically, there are other security-related flags being passed down that are good to have (stack protector and format string checking).

saiarcot895 (approved):

PR #202, opened 2023-04-24
Re-handle pre-init mux config notification after state machine init

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc/Design
- [ ] Unit test

During a config reload, mux config notification is always received before state machine init, and the state machine will not do a toggle based on the mux config after state machine init. Let's enable the state machine to re-handle the mux config notification after the state machine init and do a toggle based on the mux config.

MSFT:ADO: 17708339

#### How did you do it?
1. Add a class `AsyncEvent` that could act as an event that could execute registered callbacks if the event is notified.
2.  Define `ActiveActiveStateMachine::mWaitStateMachineInit`(of `AsyncEvent`) as a wait event to register any mux config notification received before state machine init.
3. Notify the wait event `ActiveActiveStateMachine::mWaitStateMachineInit` when activating state machine, so the mux config will be handled.

UT

### DIFF
--- a/src/common/AsyncEvent.h
+++ b/src/common/AsyncEvent.h
@@ -1,0 +1,108 @@
+/*
+ *  Copyright 2021 (c) Microsoft Corporation.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ * AsyncEvent.h
+ *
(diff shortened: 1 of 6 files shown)


yxieca:
@lolyu can you look into the CodeQL failure? looks like a compilation issue.

zjswhhh:
> @lolyu can you look into the CodeQL failure? looks like a compilation issue.

Hi @yxieca @lolyu - just to let you know I'm working on CodeQL failure. The issue seems to be hard coded line number in `sed` operating on `Makefile`. `Makefile` has recently been updated. 
https://github.com/sonic-net/sonic-linkmgrd/blob/786dc5ac48cb47e13613e4bca5f8430b19ebdb1f/.github/workflows/codeql-analysis.yml#L90-L92

yxieca (approved):

lolyu:
Hi @zjswhhh, could you please help review?

zjswhhh (approved):
lgtm. 

Since it's used for handle cached mux config, wouldn't the variable name `mWaitStateMachineInit` be too generic?

lolyu:
> lgtm.
> 
> Since it's used for handle cached mux config, wouldn't the variable name `mWaitStateMachineInit` be too generic?

@zjswhhh, I keep this generic as default route might have the same issue.

PR #216, opened 2023-08-08
[active-standby] Fix extra toggle observed in `config reload`

### LINKED ISSUE #217: [active-standby] Extra toggle observed after the standby ToR config reload.
Issues observed:
When the standby ToR config reloads, there is a possibility(less than 2%) that some ports toggle to `active`.

How to reproduce?
1. upper ToR active, lower ToR `standby`
2.  config reload the lower ToR
3. confirm if some mux ports on the lower ToR comes to `active`

lolyu:
### Description of PR

Summary:
Fixes #217

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Fix the issue that an extra mux toggle is observed when the standby ToR does a config reload.
Two reasons:
1. extra toggle due to link prober init to `unknown` when the mux is `wait`
The state transitions are:
(`unknown`, `standby`, `down`) --  activate state machine -->
(`standby`, `standby`, `down`) --> link prober `unknown` -->
(`unknown`, `standby`, `down`) --> enter mux `wait` and probe mux -->
(`unknown`, `wait`, `down`) --> port `up` -->
(`unknown`, `wait`, `up`) --> mux probe `standby` -->
(`unknown`, `standby`, `up`) --> **toggle to active** 

2. extra toggle due to the port `up` event arriving within the link prober `unknown` detection period

##### Work item tracking
- Microsoft ADO **(number only)**:  24794476

#### How did you do it?
1. when the port `up` and init the link prober state, init the link prober state to `wait` when the mux state is `wait`
2. reset the link prober state when port `up` to refresh the detection.

UT

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -247,8 +247,7 @@ void ActiveStandbyStateMachine::enterLinkProberState(CompositeState &nextState, 
 
     if (forceReset && ps(nextState) == label) {
         // only need to reset the link prober state if the state remains the same
-        link_prober::LinkProberState *currentLinkProberState = dynamic_cast<link_prober::LinkProberState *> (mLinkProberStateMachinePtr->getCurrentState());
-        currentLinkProberState->resetState();
+        mLinkProberStateMachinePtr->resetCurrentState();
     }
 
     ps(nextState) = label;
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -557,7 +557,7 @@ void ActiveStandbyStateMachine::handleStateChange(LinkStateEvent &event, link_st
             // start fresh when the link transition from Down to UP state
             // and so the link prober will initially match the MUX state
             // There is a problem with this approach as it will hide link flaps that result in lost heart-beats.
-            initLinkProberState(nextState);
+            initLinkProberState(nextState, true);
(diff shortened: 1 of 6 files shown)


src/link_manager/LinkManagerStateMachineActiveStandby.cpp
```diff
@@ -237,13 +237,19 @@ void ActiveStandbyStateMachine::setLabel(Label label) {
...
 //
-void ActiveStandbyStateMachine::enterLinkProberState(CompositeState &nextState, link_prober::LinkProberState::Label label)
+void ActiveStandbyStateMachine::enterLinkProberState(CompositeState &nextState, link_prober::LinkProberState::Label label, bool forceReset)
 {
     mLinkProberStateMachinePtr->enterState(label);
+
+    if (forceReset && ps(nextState) == label) {
+        // only need to reset the link prober state if the state remains the same
+        mLinkProberStateMachinePtr->resetCurrentState();
+    }
```
zjswhhh, line 251: @lolyu - if I understand the issue correctly, the extra toggle is trigged because the event counters are not reset when we use `enterLinkProberState` to force into a state? 

If so, why don't we just reset every time we call `enterLinkProberState`? We should still get 3 continuous heartbeat missing events before entering `LinkProberUnknon`.

lolyu: The issue is only observed during link `up`, this is to limit the change to fix this issue. We don't know what else could be caused if we reset the link prober state every time we call `enterLinkProberState`.

yxieca (approved):

zjswhhh (approved):

PR #220, opened 2023-09-20
[warmboot] config all interfaces back to `auto` if reconciliation times out 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to fix a warmboot bug that when linkmgrd reconciliation times out, some interfaces are left in `manual` mode. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

To make sure all interfaces have correct muxcable config. 

##### Work item tracking
24911432

#### How did you do it?
Run cli when timeout handler is called. 

1. Ran warmboot with healthy dualtor DUT.
2. Ran warmboot with unhealthy dualtor DUT. 

Both cases passed.

### DIFF
--- a/src/MuxManager.cpp
+++ b/src/MuxManager.cpp
@@ -570,11 +570,8 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
         MUXLOGWARNING("Reconciliation timed out after warm restart, set service to reconciled now.");
     }
 
-    uint32_t rc = system("sudo config muxcable mode auto all"); // rc > 0 means success
-
-    if (rc <= 0) {
-        MUXLOGWARNING(boost::format("config mux mode back to auto failed with return code %d") % rc);
-    }
+    int rc = system("sudo config muxcable mode auto all");
+    MUXLOGWARNING(boost::format("config mux mode back to auto completed with return code %d") % rc);
 
     mDbInterfacePtr->setWarmStartStateReconciled();
 }
--- a/src/MuxManager.cpp
+++ b/src/MuxManager.cpp
@@ -571,7 +571,10 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
     }
 
     uint32_t rc = system("sudo config muxcable mode auto all"); // rc > 0 means success
-    MUXLOGDEBUG(boost::format("config mux mode back to auto completed with return code %d") % rc);
+
+    if (rc <= 0) {
+        MUXLOGWARNING(boost::format("config mux mode back to auto failed with return code %d") % rc);
+    }
 
     mDbInterfacePtr->setWarmStartStateReconciled();
 }
--- a/src/MuxManager.cpp
+++ b/src/MuxManager.cpp
@@ -570,6 +570,9 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
         MUXLOGWARNING("Reconciliation timed out after warm restart, set service to reconciled now.");
     }
 
+    uint32_t rc = system("sudo config muxcable mode auto all"); // rc > 0 means success
+    MUXLOGDEBUG(boost::format("config mux mode back to auto completed with return code %d") % rc);
+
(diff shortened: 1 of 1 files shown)


src/MuxManager.cpp
```diff
@@ -570,6 +570,9 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
         MUXLOGWARNING("Reconciliation timed out after warm restart, set service to reconciled now.");
     }
 
+    uint32_t rc = system("sudo config muxcable mode auto all"); // rc > 0 means success
```
vaibhavhd, line 573: Do you not want to handle rc, at least for NOTICE vs WARNING logs? If config to auto fails, do you still want to conclude to reconciled state?

vaibhavhd, line 573: I think we can add a UT to ensure when reconciliation is completed the config is indeed set to auto?

zjswhhh: > Do you not want to handle rc, at least for NOTICE vs WARNING logs? If config to auto fails, do you still want to conclude to reconciled state?

What will happen if linkmgrd never enter "reconciled" state?

zjswhhh: It's a bit hard, `linkmgrd` UTs doesn't use mock DBs.

vaibhavhd: The finalizer script waits on linkmgrd to reconcile: https://github.com/sonic-net/sonic-buildimage/blob/7cd9f132535a90c2080cccfc418b976e471f69b5/files/image_config/warmboot-finalizer/finalize-warmboot.sh#L13C31-L13C31

I can't think of a way to fix this failed to reconcile state. The best I can think of is if the reconciliation did not happen within 5 minutes then we consider warmboot failed, and a log is generated from finalizer.
There is no alert generated today (AFAIK), but I think we should generate alert when such a log is created.

yxieca: I think a warning on failure woud be nice.

src/MuxManager.cpp
```diff
@@ -570,6 +570,12 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
...
+
+    if (rc <= 0) {
```
saiarcot895, line 575: I don't think the return code would be negative (unless it is -1 because the bash process couldn't be created)? Based on the man page, the return value will be similar to a wait status return value.

vaibhavhd, line 575: Are we sure that rc>0 is a success case? Since here system command is executed, and in Linux system return code > 0 is error.

saiarcot895: The data type is also `uint32_t`, so this is logically incorrect.

zjswhhh: I was referring to the case when child process could not be created (rc==0 or rc==-1). And log was intended for debugging purpose only. 

Guess the wording is misleading. Will update. 

reference: https://man7.org/linux/man-pages/man3/system.3.html

zjswhhh: > The data type is also `uint32_t`, so this is logically incorrect.

Oh good catch.

src/MuxManager.cpp
```diff
@@ -570,6 +570,9 @@ void MuxManager::handleWarmRestartReconciliationTimeout(const boost::system::err
         MUXLOGWARNING("Reconciliation timed out after warm restart, set service to reconciled now.");
     }
 
+    int rc = system("sudo config muxcable mode auto all");
+    MUXLOGWARNING(boost::format("config mux mode back to auto completed with return code %d") % rc);
```
vaibhavhd, line 574: With this change, there will always be a WARNING regardless of successful mux switch or not.

zjswhhh: Which is fine, linkmgrd's default log level is `warning`.

vaibhavhd (approved):

saiarcot895 (approved):

PR #225, opened 2023-12-12
[active-standby] Fix `show mux status` inconsistency introduced by orchagent rollback 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
This is to fix the `show mux status` inconsistency introduced by orchagent roll back. 

In mux port state machine design, linkmgrd honors hardware state for active-standby ports, and never intends to trigger a secondary toggle when everything is healthy. But after we introduce orchagent rollback, `show mux status` can return unmatched APP_DB and STATE_DB entries for this, which blocks upgrade. 

Hence, submitting this PR as a workaround.  

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

##### Work item tracking
26136887

### DIFF
--- a/test/MuxManagerTest.cpp
+++ b/test/MuxManagerTest.cpp
@@ -924,7 +924,7 @@ INSTANTIATE_TEST_CASE_P(
     MuxState,
     GetMuxStateTest,
     ::testing::Values(
-        std::make_tuple("active", mux_state::MuxState::Label::Active),
+        std::make_tuple("active", mux_state::MuxState::Label::Wait),
         std::make_tuple("standby", mux_state::MuxState::Label::Wait),
         std::make_tuple("unknown", mux_state::MuxState::Label::Wait),
         std::make_tuple("error", mux_state::MuxState::Label::Wait)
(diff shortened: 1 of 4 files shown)


test/MuxManagerTest.cpp
```diff
@@ -924,7 +924,7 @@ INSTANTIATE_TEST_CASE_P(
     MuxState,
     GetMuxStateTest,
     ::testing::Values(
-        std::make_tuple("active", mux_state::MuxState::Label::Active),
+        std::make_tuple("active", mux_state::MuxState::Label::Wait),
```
prsunny, line 927: Can you please help me understand the difference between the Wait and Active?

zjswhhh: `Wait` is when linkmgrd writes to APP_DB, it will "wait" for the toggle/muxprobe result, so it will have the latest hw state. 

`Active` is simply as it implies. 

The unit test here is to validate how an **active** port reacts to STATE_DB value == "active". It used to be no-op, so linkmgrd state remains "Active", but now we are enforcing another toggle, so it will be "Wait".

prsunny: ok, so can I assume `Standby `is already handled in this flow? Since you added test for only `Active`

prsunny (dismissed):

test/MuxManagerTest.cpp
zjswhhh: > ok, so can I assume `Standby `is already handled in this flow? Since you added test for only `Active`

Yes.

prsunny (approved):

lolyu (approved):

prsunny:
@zjswhhh , would you raise a PR for 202205?

StormLiangMS:
@zjswhhh cherry pick conflict, could you help to raise separate PR for 202305?

PR #235, opened 2024-01-11
[active-standby] Probe the link in suspend timeout

### LINKED ISSUE #208: [active-standby] Indefinitely heartbeat suspending on active side if standby peer doesn't switch
Issue introduced by https://github.com/sonic-net/sonic-linkmgrd/commit/faceb93c1953321580771596be70765b131b03f7

lolyu:
### Description of PR

Summary:
Fixes #208

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Fix the issue:
After the link probe `unknown`, the `standby` ToR fails to take the link.
When the link recovers, the `active` ToR is stuck in heartbeat suspension.

##### Work item tracking
- Microsoft ADO **(number only)**: 26281338

#### How did you do it?
Let's probe the link in suspension timeout if current state is (unknown, active, up).
Kudos to @zjswhhh to point out we should force sending the heartbeat in link detect.

UT and verify on testbed.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -493,6 +493,7 @@ void ActiveStandbyStateMachine::handleStateChange(LinkProberEvent &event, link_p
 
     if (ps(mCompositeState) != link_prober::LinkProberState::Unknown) {
         mResumeTxFnPtr();
+        mUnknownActiveUpBackoffFactor = 1;
     }
 
     updateMuxLinkmgrState();
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -366,6 +366,9 @@ void ActiveStandbyStateMachine::handleSwssBladeIpv4AddressUpdate(boost::asio::ip
             mProbePeerTorFnPtr = boost::bind(
                 &link_prober::LinkProber::probePeerTor, mLinkProberPtr.get()
             );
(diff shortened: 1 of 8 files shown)


src/link_manager/LinkManagerStateMachineActiveStandby.cpp
```diff
@@ -829,6 +832,8 @@ void ActiveStandbyStateMachine::handleSuspendTimerExpiry()
         CompositeState currState = mCompositeState;
         enterMuxWaitState(mCompositeState);
         LOGINFO_MUX_STATE_TRANSITION(mMuxPortConfig.getPortName(), currState, mCompositeState);
+        // detect if link/server recovers
+        mDetectLinkFnPtr();
```
zjswhhh, line 836: Let's send the heartbeats first, then `enterMuxWaitState` (probe mux state).

zjswhhh: Also is there a way to make sure all 3 `sendHeartbeat` are dispatched and executed before probing mux state.

zjswhhh (changes requested):

src/link_manager/LinkManagerStateMachineActiveStandby.cpp
lolyu: Thanks, force the sending in link detect.

liushilongbuaa:
Sorry for the auto cherry pick issue. 202205/202211 have conflict. 202305/202311 don't have conflict.
I just fixed the automation issue.

zjswhhh (approved):

PR #245, opened 2024-03-13
[active-standby] reset mux probing backoff factor in link flaps

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to address consistency in PXE boot scenario. In PXE boot, there is no ICMP heartbeats. During booting up, `standby` might switch to `active`. Mux probing is now the only way to find out mux toggles. But it's also expected to see mux probing fail when link is down, backoff factor can already be maximum when link is up. 

If we don't reset backoff factor, it might take ~40s for peer to correct the inconsistency. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

##### Work item tracking
467321367


theasianpianist (dismissed):

lolyu (dismissed):

theasianpianist (approved):

Ndancejic (approved):

prsunny:
@zjswhhh , need this backported. Let me know if you've concerns.

PR #257, opened 2024-06-24
Fix state machine handler race conditions introduced by `strand::wrap`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

##### Work item tracking


lolyu:
cannot repro the UT failure on local device, will investigate soon.

yxieca:
@lolyu please look into the checker failures.

test/FakeMuxPort.cpp
```diff
@@ -36,7 +37,8 @@ FakeMuxPort::FakeMuxPort(
     std::string &portName,
     uint16_t serverId,
     boost::asio::io_service &ioService,
-    common::MuxPortConfig::PortCableType portCableType
+    common::MuxPortConfig::PortCableType portCableType,
+    bool useFakeLinkManagerStateMachine
```
zjswhhh, line 41: Why do we need a fake state machine? Is it relevant in this PR? When do we use it?

src/MuxPort.cpp
```diff
@@ -96,12 +96,11 @@ void MuxPort::handleBladeIpv4AddressUpdate(boost::asio::ip::address address)
 {
     MUXLOGDEBUG(boost::format("port: %s") % mMuxPortConfig.getPortName());
 
-    boost::asio::io_service &ioService = mStrand.context();
-    ioService.post(mStrand.wrap(boost::bind(
+    boost::asio::post(mStrand, boost::bind(
```
zjswhhh, line 99: `wrap()` occurs in other places too, do we plan to remove them all? If yes, do you want to put it in the same PR?

zjswhhh (commented):
When will the

zjswhhh (commented):
Please add PR description.

test/FakeMuxPort.cpp
lolyu: when we want to test the state machine handler order, like the `MuxPortTest`

src/MuxPort.cpp
lolyu: This PR only removes the `wrap` in the `MuxPort` class.

lolyu:
> Please add PR description.

the PR description is there

zjswhhh (approved):
lgtm

PR #261, opened 2024-07-01
[active-standby] Fix the oscillation logic

### LINKED ISSUE #253: [active-standby] Unexpected toggle due to the oscillation logic
There is an unexpected toggle observed due to the oscillation.
The background is that, the oscillation timer is triggered if the mux is in (wait, active, up), and when it is expired, if the mux is still in (wait, active, up), linkmgrd will toggle to standby.
But as (wait, active, up) is an intermediate state between the toggle path from standby to active, so the oscillation timer could be triggered from a legit toggle from standby to active. Also, when the timer is expired in the period when mux is toggling from standby to active and the mux is in (wait, active, up), the mux will toggle back to standby. The extra toggle from the oscillation is disruptive here.

The whole story goes like the following:
image

lolyu:
### Description of PR

Summary:
Fixes #253

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

Fix the unexpected toggle introduced by the oscillation logic.

##### Work item tracking
- Microsoft ADO **(number only)**: 28397786

#### How did you do it?
The oscillation is introduced to allow the active side to toggle to standby if no heartbeat is received.
The workflow is described as the following:
```
(wait, active, up) ---> set oscillation timer [1]
...
(wait, active, up) ---> (wait, wait, up) [2]
(wait, wait, up) ---> (wait, active, up)
...
(wait, active, up) <--- oscillation timer expires, toggle to standby [3]
```
[1]: the ToR enters `(wait, active, up)`, no heartbeats received, oscillation timer is set.
[2]: the ToR consistently probes the mux status, and transits between `(wait, active, up)` and `(wait, wait, up)`.
[3]: when the oscillation timer expires and the ToR is `(wait, active, up)`, make the toggle to standby request.

We need to ensure that, the ToR is only allowed to transits between `(wait, active, up)` and `(wait, wait, up)` during [2]. So any link prober active/standby, mux standby, or link down events should cancel the oscillation.

UT

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveStandby.cpp
@@ -61,8 +61,6 @@ ActiveStandbyStateMachine::ActiveStandbyStateMachine(
     mMuxStateMachine.setWaitStateCause(mux_state::WaitState::WaitStateCause::SwssUpdate);
     mMuxPortPtr->setMuxLinkmgrState(mLabel);
     initializeTransitionFunctionTable();
-
-    mOscillationTimer.expires_from_now(boost::posix_time::seconds(1)); 
 }
 
 //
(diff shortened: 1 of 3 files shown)


src/link_manager/LinkManagerStateMachineActiveStandby.cpp
```diff
@@ -476,12 +474,20 @@ void ActiveStandbyStateMachine::handleStateChange(LinkProberEvent &event, link_p
             mMuxPortPtr->postLinkProberMetricsEvent(link_manager::ActiveStandbyStateMachine::LinkProberMetrics::LinkProberActiveStart);
 
             mStandbyUnknownUpCount = 0;
+
+            if (mOscillationTimerAlive) {
+                cancelOscillationTimer();
+            }
```
yxieca, line 480: Not to add to this PR. Suggest move the check into cancellation helper function, and rename the function to tryCancel......()

yxieca (approved):

src/link_manager/LinkManagerStateMachineActiveStandby.cpp
lolyu: Addressed, thanks!

yxieca (approved):

yxieca:
@lolyu do we not need this change in 202311?

zjswhhh:
> @lolyu do we not need this change in 202311?

Was the issue found in production or nightly?

PR #266, opened 2024-08-07
[link prober] print out error code message when `sendHeartbeat` fails

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
Print out error code message when `sendHeartbeat` fails.

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

##### Work item tracking

### DIFF
--- a/src/link_prober/LinkProber.cpp
+++ b/src/link_prober/LinkProber.cpp
@@ -333,7 +333,7 @@ void LinkProber::sendHeartbeat(bool forceSend)
         mStream.write_some(boost::asio::buffer(mTxBuffer.data(), mTxPacketSize), errorCode);
 
         if (errorCode) {
-            MUXLOGTRACE(mMuxPortConfig.getPortName() + ": Failed to send heartbeat!");
+            MUXLOGTRACE(mMuxPortConfig.getPortName() + ": Failed to send heartbeat! Error code: " + errorCode.message());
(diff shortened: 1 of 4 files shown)


src/link_prober/LinkProber.cpp
```diff
@@ -333,7 +333,7 @@ void LinkProber::sendHeartbeat(bool forceSend)
         mStream.write_some(boost::asio::buffer(mTxBuffer.data(), mTxPacketSize), errorCode);
 
         if (errorCode) {
-            MUXLOGTRACE(mMuxPortConfig.getPortName() + ": Failed to send heartbeat!");
+            MUXLOGTRACE(mMuxPortConfig.getPortName() + ": Failed to send heartbeat! Error code: " + errorCode.message());
```
lolyu, line 336: what's the output of `errorCode.message()`?

lolyu (commented):
do we also need to catch the I/O error in the receive?

src/link_prober/LinkProber.cpp
zjswhhh: A string, i.e. for `boost::system::errc::bad_file_descriptor` it will return "Bad file descriptor"

lolyu (approved):
LGTM

PR #289, opened 2025-04-07
[active-standby] shutdown link prober when starting as isolated

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
Noticed that if server does serving during LT0 ugprade, link flaps on both sides' ToR, T0 can't always take over as `active` after link is up. Cause in some cases the `shutdownTx` was skipped on LT0 due to state machine inactivated or `tsa_enabled` made mux mode `standby`. Logs attached in internal ADO. 

Adding a similar fix to https://github.com/sonic-net/sonic-linkmgrd/pull/130 for active-standby as well. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

##### Work item tracking
32099536

UTs


prsunny (approved):
lgtm

PR #291, opened 2025-04-29
[active-active] Fix mux not toggle to active in default route flapping

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

If the default route flaps faster than the link prober detection, the mux status remains `standby` even after default route `ok`:

```
(active, active, up) ----> default route n/a, toggle to standby, suspend sending heartbeat ---->
(active, standby, up) ----> default route ok, resume heartbeat sending ---->
(active, standby, up) ----> link prober active, noop ---->
(active, standby, up)
```

##### Work item tracking
- Microsoft ADO **(number only)**: 32625971

#### How did you do it?
Reset the link prober state in this case:
```
(active, active, up) ----> default route n/a, toggle to standby, suspend sending heartbeat ---->
(active, standby, up) ----> default route ok, resume heartbeat sending, reset link prober state ---->
(unknown, standby, up) ----> link prober active, toggle to active ---->
(active, active, up)
```

UT and on DUT.

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -1294,6 +1294,7 @@ void ActiveActiveStateMachine::handleDefaultRouteStateNotification(const Default
 {
     MUXLOGWARNING(boost::format("%s: default route state %d") % mMuxPortConfig.getPortName() % static_cast<int>(routeState));
 
+    DefaultRoute oldRouteState = mDefaultRouteState;
     mDefaultRouteState = routeState;
     shutdownOrRestartLinkProberOnDefaultRoute();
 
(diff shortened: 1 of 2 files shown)


src/link_manager/LinkManagerStateMachineActiveActive.cpp
```diff
@@ -1307,6 +1308,19 @@ void ActiveActiveStateMachine::handleDefaultRouteStateNotification(const Default
...
+        ps(mCompositeState) == link_prober::LinkProberState::Label::Active &&
+        mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Auto) {
+        // Let's reset the link prober state if the default route flaps faster than
+        // the link prober detection.
+        MUXLOGWARNING(
+            boost::format("%s: reset link prober state, current mux state '%s' ") %
+            mMuxPortConfig.getPortName() %
+            mMuxStateName[ms(mCompositeState)]
+        );
+        initLinkProberState(mCompositeState);
```
yxieca, line 1322: This change will fix the sequence you target to fix. Is there a downside for this change? What if the link prober is already having a notification to linkmgrd outstanding? In such case, would reset cause an extra flap?

zjswhhh (commented):
Hi @lolyu - why would default route flap? Was it a test failure or production scenario? I assume for default route to recover to `ok` we need BGP sessions to be unshut and converge, that will take longer than 300ms? 

As for the fix itself, feel like it's too ad-hoc. What about we just do a link prober state reset every time we resume link prober sending? Same logic as link up.

zjswhhh (approved):
The checkers are specific to this case, hence shouldn't be impactful for the state machines. Approve for now. Please consider my comment above though.

zjswhhh:
Why is this needed for 202411 if it's not a production issue?

src/link_manager/LinkManagerStateMachineActiveActive.cpp
lolyu: if link prober has notified the state machine, its link prober state should be changed instead of `Active`.

lolyu:
> Why is this needed for 202411 if it's not a production issue? Also how is it even possible to cherry-pick to 202411.

This issue is observed in 202411 qual, and it is needed to unblock the Cisco unboard.

PR #323, opened 2025-12-16
Fix periodic sync recovery when mux config mode differs from hardware state.

rajkumar1-arista:
### Description of PR
When mux is configured in manual mode (`active/standby`), periodic sync detects hardware drift but fails to trigger resync. After a successful toggle, `mLastMuxStateNotification` matches the expected state, so the existing check passes and no resync occurs even when probe detects drift. Fix by adding an `else-if` branch that checks `mLastMuxProbeNotification` when the notification came from a probe, triggering resync if hardware state differs from expected.

Summary:
Fixes # (https://github.com/aristanetworks/sonic-qual.msft/issues/947)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Doc/Design
- [ ] Unit test

When mux is configured in manual mode (`active/standby`), periodic sync detects hardware drift but fails to trigger resync.

##### Work item tracking
- Microsoft ADO **(number only)**: 35946943

#### How did you do it?
By adding an `else-i`f branch that checks `mLastMuxProbeNotification` when the notification came from a probe, triggering resync if hardware state differs from expected.

Ran `test_mux_forwarding_state_consistency`

### DIFF
--- a/src/link_manager/LinkManagerStateMachineActiveActive.cpp
+++ b/src/link_manager/LinkManagerStateMachineActiveActive.cpp
@@ -784,6 +784,10 @@ void ActiveActiveStateMachine::LinkProberActiveMuxActiveLinkUpTransitionFunction
     if (mMuxPortConfig.getMode() == common::MuxPortConfig::Mode::Standby) {
         if (mLastMuxStateNotification != mux_state::MuxState::Label::Standby) {
             switchMuxState(nextState, mux_state::MuxState::Label::Standby, true);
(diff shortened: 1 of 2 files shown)


lolyu (commented):
So the root cause that the linkmgrd doesn't correct the mux after detecting hardware drift is the mux config to active/standby, right?

lolyu (approved):
LGTM, thanks for the fix; please address Jing's comment.

zjswhhh (approved):

PR #339, opened 2026-04-16
Simplify CodeQL: download pre-built swss-common from CI artifacts

yxieca:
#### What I did
Simplified the CodeQL workflow by downloading pre-built sonic-swss-common packages from Azure DevOps CI instead of building from source.

#### How I did it
- **Replaced** the `build-swss-common` step (which cloned and built swss-common from source) with an `install-swss-common` step that downloads the official pre-built `libswsscommon-dev` deb from Azure DevOps pipeline 9 via the public REST API
- This is the same approach that `azure-pipelines.yml` already uses via `DownloadPipelineArtifact@2`
- Downloads artifacts for the **matching branch** (PR target or push branch)
- The pre-built deb contains a complete `cfg_schema.h` with all `CFG_*` macros generated from YANG models
- **Removed** the `generate-cfg-schema` step (regex-based YANG parser) since it is no longer needed
- **Upgraded** boost from 1.74 (`libboost-all-dev`) to 1.83 (`ppa:mhier/libboost-latest`) to match the pre-built deb, which was built in the sonic-slave-bookworm container against boost 1.83
- **Added graceful fallback**: if the Azure DevOps download fails, CodeQL still runs static analysis (skips the build step)

#### Why this is needed
The previous workflow built swss-common from source and used a regex script to generate `cfg_schema.h` from YANG files. Downloading the official CI artifacts is simpler, faster, and always correct.

#### How to verify it
- CodeQL CI should pass (both compilation and analysis)
- The `install-swss-common` step logs the build ID, branch, and macro count
- If Azure DevOps is unreachable, the workflow degrades gracefully to static-only analysis

#### Description for the changelog
Simplify CodeQL workflow by downloading pre-built swss-common from Azure DevOps CI instead of building from source, and upgrade boost to 1.83 to match.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -76,55 +76,12 @@ jobs:
         git clone https://github.com/sonic-net/sonic-swss-common
         pushd sonic-swss-common
         ./autogen.sh
-        git checkout ${GITHUB_BASE_REF:-master}
-        git reset HEAD --hard
         dpkg-buildpackage -rfakeroot -us -uc -b -Pnoyangmod,nopython2 -j$(nproc)
         popd
         dpkg-deb -x libswsscommon_${SWSSCOMMON_VER}_amd64.deb $(dirname $GITHUB_WORKSPACE)
         dpkg-deb -x libswsscommon-dev_${SWSSCOMMON_VER}_amd64.deb $(dirname $GITHUB_WORKSPACE)
       env:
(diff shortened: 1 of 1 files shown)


lolyu (approved):
### ✅ Approved

Clean simplification — download pre-built artifact instead of building from source. Graceful fallback, cfg_schema.h validation, conditional build step. LGTM 🚀

PR #345, opened 2026-06-15
[ci]: Install libyang3 for linkmgrd builds

zjswhhh:
## Why I did it

linkmgrd CI currently assumes the common-lib artifact contains `libyang_*.deb`, but the current SONiC common-lib artifact publishes `libyang3_*.deb` and `libyang-dev_*.deb`. This causes Azure to download zero packages and then fail with `dpkg --install needs at least one package archive file argument`.

CodeQL also downloads `libswsscommon` without extracting the matching SONiC `libyang3`, so the link step cannot resolve `libyang.so.3`.

## How I did it

- Update Azure artifact patterns to download `libyang3_*.deb` and `libyang-dev_*.deb`.
- Make Azure fail with a clear message if no libyang packages are downloaded.
- Extract matching common-lib `libyang3` packages in CodeQL before building against downloaded `libswsscommon`.
- Add `-Wl,-rpath-link` for the extracted library directory so CodeQL's linker can resolve `libyang.so.3`.

## How to verify it

- Run the Azure `sonic-linkmgrd` pipeline.
- Run the GitHub CodeQL workflow.
- Confirm the libyang install step finds packages and CodeQL no longer fails with unresolved `ly_*` symbols from `libswsscommon`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -88,11 +88,17 @@ jobs:
       path: $(Build.ArtifactStagingDirectory)/download
       artifact: common-lib
       patterns: |
-        target/debs/bookworm/libyang_*.deb
+        target/debs/bookworm/libyang3_*.deb
+        target/debs/bookworm/libyang-dev_*.deb
     displayName: "Download libyang from common lib"
   - script: |
       set -ex
-      sudo dpkg -i $(find ./download -name *.deb)
+      mapfile -t libyang_debs < <(find ./download -name '*.deb' -print)
+      if (( ${#libyang_debs[@]} == 0 )); then
+        echo "No libyang packages found in downloaded common-lib artifact"
+        exit 1
+      fi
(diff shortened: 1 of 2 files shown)


Ndancejic (approved):

azure-pipelines.yml
```diff
@@ -88,11 +88,17 @@ jobs:
       path: $(Build.ArtifactStagingDirectory)/download
       artifact: common-lib
       patterns: |
-        target/debs/bookworm/libyang_*.deb
+        target/debs/bookworm/libyang3_*.deb
+        target/debs/bookworm/libyang-dev_*.deb
```
theasianpianist, line 92: Nit: suggest pattern `libyang-dev_3*.deb` for the dev headers package instead. Shouldn't affect functionality since libyang1 is no longer being built anyways

theasianpianist (approved):

